PR #11, opened 2018-03-21
Use webpack for builds

olivereisenhut:
Why
===
onsidering using Vue or an other MVVM you should change to a bundler like webpack. Webpack provides [loaders](https://webpack.js.org/loaders/) which transform your required files if they aren't pure JS. It also handles other files like SCSS and so on.

How
=== 
There are two bundles. One for the main which runs on node an one for the renderer. They share a config in the `webpack/webpack.common.js` folder and extend have their own config to extend their behaviour: `webpack/webpack.main.js` and `webpack/webpack.renderer.js`. Finally all the configs get merged an loaded in `webpack.config.js`

Because there there remain just two scripts some paths have to be adjusted which depended on the old folder structure.


olivereisenhut:
Vscode setup needs some more work

PR #16, opened 2018-04-29
Add Calculator feature

khanhas:
First, I want to give huge thank to @oliverschwendener for create this beautiful app. It looks so good and works so fast. And I want to expand it more.
Calculating stuff plays big part in my daily works. But opening up Calculator in Windows is too slow, Cortana sometimes refuses to recognize my formula and returns web search instead.
So I think implement calculator right in this launcher solves my problem and probably many user too.


oliverschwendener:
Thank you very much for your contribution, this a very handy feature! I am very happy to include your changes into my project. But before merging I want to suggest a few improvements:

* You should have unit tests that test your classes `Calculator` and `CalculatorInputValidator`. For an example take a look at:
    * `src/tests/unit/web-url-searcher.test.ts`.
    * `src/tests/unit/input-validators/web-url-input-validators.test.ts`
* If the userInput is "sqrt()" the result is a 'stringified' function, here a screenshot:
image Please check in the `isValidForSearchResults` method if the result is a number. I would suggest using the `isNan()` function.
* Please make a section in the README.md which quickly explains your feature (for example like the 'Execute commandline tools' section)

khanhas:
I added all the thing you mentioned. Man, I learned a lot of new thing today.
Anyway, there's still a small problem that Calculator result still be executed and throw error when I hit enter.

khanhas:
Do you know what can I do to fix this "coverage" error?
https://coveralls.io/builds/16765666/source?filename=src%2Fts%2Finput-validators%2Fcalculator-input-validator.ts#L18

oliverschwendener:
@khanhas With this line of code you are checking if the input is of type object am I right?
``` typescript
if (typeof(input) !== "object") {
    return false;
}
```

In your test cases...
``` typescript
it("should return false when passing in an invalid argument", (): void => {
    const invalidInputs = [
        "s",
        "kg",
        "log()",
        "1 blackhole to mg",
    ];

    for (const invalidInput of invalidInputs) {
        const actual = validator.isValidForSearchResults(invalidInput);
        expect(actual).toBe(false);
    }
});
```
...you are not testing case where the result of your input is type of object. Therefore the if-statement is not covered in your tests. If you extend your tests with a case where the result of your input is type of object the if-statement will be covered in your tests and the code coverage should not be decreased.

khanhas:
Oh I get it. It should be OK if I remove that object type check tho, since `math.eval` return either a number or an object. It's not possible for it to returns anything else.

oliverschwendener:
Sounds good to me.

PR #18, opened 2018-05-01
WIP: Refactor config for better testability

olivereisenhut:
Split up the config into multiple files so it's easier to test them


khanhas:
While you're working on config, is there any chance you can make `reload ueli` command to reload config file?

PR #24, opened 2018-05-05
Sort results by frequently used file path

khanhas:
I want to propose an improvement on how search engine sorting results. 
As I inspected your Fuse config, it sorts files and settings based on their name and tags, ranking results with `score`. Lower `score` = closer match to user's query. Most of the time, it works well, I only need to type in 5, 6 characters to reach correct result. 
To decrease typing and browsing search result time for daily programs/settings, search engine should "learn" over time what programs/settings are mostly used and put them first.
This idea is still rough along with my naive attempt to make it work. Feel free to criticize, comment or suggest.


olivereisenhut:
I would suggest to save the rating file rather in the app’s directory than in the home direcotory.

oliverschwendener:
Thanks for your PR! I have a few suggestions for improvement.
1. Rename the class WeightManager to ExecutionCountManager or something similar because it is only counting how often something was executed, am I right?
2. Your test is in the "unit" folder but creates a file on the file system. Therefore it should be in the "integration" folder which holds all integration tests.
3. The call to this method: `WeightManager.addWeightScore(executionArgument);` should be on a higher level (I suggest in the execution service). You could extend the interface "Executor.ts" by a method/property "LogExecution" or something similar which controls if an executor wants to log the execution.
4. In my opinion the weight is to "aggressive". After I executed a program only once it already shows up at the top of the search results. I suggest to add some sort of multiplier, to reduce the aggressiveness of the weight in the search results.
5. Why is everything in the WeightManager class static?

oliverschwendener:
Nope, code coverage is only checked for unit tests.

Here are some suggestions how you could have better tests (and coverage):

The class "CountManager" should not depend on the file system and should only contain logic. Create another class and interface which reads/writes from/to the file system. In the real application you inject this class into the count manager and in the tests you can inject a "fake" repository to simulate the access to the file system. This way you can cover all logical code in unit tests. Do you know what I mean?

Furthermore, extend the interface `Executor` by a method `logExecution(): boolean`.

``` typescript
export interface Executor {
    execute(executionArgument: string): void;
    hideAfterExecution(): boolean;
    resetUserInputAfterExecution(): boolean;
    logExecution(): boolean;
}
```

Then the class "FilePathExecutor", "WindowsSettingsExecutor" and "MacOsSettingsExecutor" return `true` when calling this method/property. All other executors should return `false`.

``` typescript
public logExecution(): boolean {
    return true;
}
```

In the "ExecutionService" class instead of this:

``` typescript
if (executorName.match(/FilePathExecutor|WindowsSettingsExecutor|MacOsSettingsExecutor/)) {
    this.logExecute(executionArgument);
}
```

... do this:

``` typescript
if (combi.executor.logExecution()) {
    countManager.addCount(executionArgument);
}
```

So each executor can control if its execution should be counted.

Thank you very much for your contribution! I Really appreciate your effort!

khanhas:
@oliverschwendener Thank you for your really detail instruction. Most of the time I don't know how can to do it right and optimal. So I extremely appreciate your patience with this.  
Something that still confuse me that when I should use `static` modifier in class's methods, when I shouldn't. 
Only difference I can notice between them is that with static methods when I `import { Something } from "./something"`, I can use `Something.method` directly. And I don't have to initialize a whole object with `new` which seems wasting resource for one simple method execution.

oliverschwendener:
Feature has been implemented.

PR #31, opened 2018-06-06
Search and autocomplete enviroment/custom variable path

khanhas:
Since we've already have shorthand for programs execution, how about shorthand for access directory?
Quickly gets into a deep directory instead of starting from root every time.

Get environment variables via `process.env` object, validate all of its fields to get only variable that has a valid path as value.
I also added a field in config allow user to add more variables themselves.

When hitting tabs to autocomplete, variable value will replace user input. Then user can continue to search, autocomplete or open deeper level directory.

Demo: https://i.imgur.com/9Oa4iSd.mp4


oliverschwendener:
This i a very nice idea, thank you for your contribution again ;) ! The code looks also very good. But I wonder if there is a way to merge this feature with the "custom commands" feature. It's almost the same. You have a shortcut (for example: `data`) which leads to an executionArgument (for example: `C:\Data`). We would have to remodel the autocompletion section in `main.ts` to autocomplete all executionArguments that are a valid filepath and additionally we could also pass the environment variables from `process.env` to custom commands or even a separate plugin which covers only the envrionment variables.

oliverschwendener:
Added better autocompletion in 47f1e95

khanhas:
Ok, using it as plugin seems like a better idea.
Do you think we should log autocomplete execution arguments like we do when executing them?
Because in my experience with it, if we don't have prefix to discriminate field of things, it searches pretty much all of them, commands, files, variables and taking quite some steps to get what I want and use frequently.

oliverschwendener:
Only log on execution and not on autocompletion.

oliverschwendener:
I'll merge later today. I will take care of the `!` in custom commands.

oliverschwendener:
@khanhas 

Over the weekend I thought about your PR. The idea is really good, but your implementation does not fit with my idea. I think we can merge your "directory variable" feature with the existing "custom commands" feature. I remodelled the custom command plugin that we can specify custom directories via custom commands like this: 

``` json
{
    "name": "Data",
    "executionArgument": "C:\\Data"
}
```

If you want to start a command line tool inside ueli:

``` json
{
    "name": "ping",
    "executionArgument": ">ping 8.8.8.8 -t"
}
```

If you want to start a command line tool outside of ueli:

``` json
{
    "name": "ping",
    "executionArgument": "!start ping 8.8.8.8 -t"
}
```

Additionally you can activate the search for environment variables (which is a separate plugin) via config options (default is `false`):

``` json
{
    "searchEnvironmentVariables": true
}
```
image

I won't merge your PR because I don't want to remodel your code to fit my idea. But I will mention you in the release notes because you were the one who came up with this great idea.

PR #67, opened 2018-11-22
Better icon for Windows Taskbar

mkkeck:
See this:
windows-icon


oliverschwendener:
I see your point, the current icon can't be seen on white background. Maybe we should just use a black variant of the current icon instead of adding an outline.

oliverschwendener:
Or maybe use an icon with a solid black background

mkkeck:
I think the outlined icon is much more flexible and looks better than one with solid background.
And why only for Windows 7?

mkkeck:
I think, it does not make sense to post pull request here, if no one of this project want my work (perhaps it's not good enough or to many changes at once)
So I close this.

oliverschwendener:
I really appreciate your effort and your interest in contributing to this project. But in my opinion the icon with the black outline just doesn't look good.

olivereisenhut:
Hey @mkkeck. It is very hard for the maintainers of a project to process a big PR. There are a lot of changes without any context why something has changed an so on. I would suggest to create issues before creating any bigger PR. In those issues you can document the whats and whys. This helps to give context to changes and maybe it comes out that some changes are not even needed.

Just some thoughts from my site.

PR #149, opened 2019-06-02
refactor: use async-await instead of promises

cliffordfajardo:
# Changes
**I realize that this is a pretty big Pull request, however:**
- it adds no new features
- is simply a refactor so the code works the same way
- tested on Windows & Mac against the `dev` branch

**1. Refactor `Promise` usage in favor of `async` functions where possible.**

<details>
Async/await allows us to write promise-based code as if it were synchronous without blocking the main thread.

The only places I didn't convert to async functions are the {translation, mdfind, everything} plugins. I had trouble using async/await with `setTimeout` in those 3 plugins. 

Besides the syntactic sugar, there are many strong justifications for using async/await. Here are a few:

Resource links:

- [Asynchronous stack traces: why await beats Promise.then](https://mathiasbynens.be/notes/async-stack-traces)

- [Async functions - making promises friendly](https://developers.google.com/web/fundamentals/primers/async-functions)

- [Async/await](https://javascript.info/async-await)
</details>

**2. Updated sub-dependencies with `yarn upgrade`**

<details>
This fixed a npm/yarn install failure on node 12.20 on Windows.

The error that was happening on the Windows operating system during `yarn` or `npm install` was the following:

```
upath@1.0.4: the engine "node" is incompatible with this module.
```

This would preventing installing node_modules for the project. A yarn upgrade fixed the [issue](https://github.com/pusher/react-slack-clone/issues/69)

Lastly, I added the package `"@types/node": "^12.0.4"` to get more intellisense completions inside the project (I'm using VS Code too).
</details>

**3. Updated `tsconfig.json`'s bundle target options to `ESNEXT`**
Since we're running the latest version of Node we don't need to transpile for older runtimes. The following is nothing too special BUT the bundle size from 4.33mb to 4.30mb. I bet there are more optimizations that could be done in the future.

**4. Small miscellaneous changes**

<details>

- added precommit hook that runs `tslint` check before allowing the user to commit. If tslint fails, then the git commit will exit and prompt the user to fix their changes. I learned about my `tslint` errors only until the CI failed. Adding this will save users time.

- I added a few type declarations where they were missing, mostly to silence the typescript compiler 😄 

<details>

I'm also in your slack channel. Feel free to send me time ranges for when you're mostly available. Feel free to add as many comments/questions to the PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,9019 @@
+{
+  "name": "ueli",
+  "version": "7.2.5",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "7zip-bin": {
+      "version": "4.1.0",
+      "resolved": "https://registry.npmjs.org/7zip-bin/-/7zip-bin-4.1.0.tgz",
+      "integrity": "sha512-AsnBZN3a8/JcNt+KPkGGODaA4c7l3W5+WpeKgGSbstSLxqWtTXqd1ieJGBQ8IFCtRg8DmmKUcSkIkUc0A4p3YA==",
+      "dev": true
+    },
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "dev": true,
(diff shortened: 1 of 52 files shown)


oliverschwendener:
Hey @cliffordfajardo, I'm sorry that it took so long until I could review your PR. Thank you for contributing to this project!

I see the main benefits of using async/await instead of promises and I totally agree with you. However this PR is quite huge and contains different kinds of changes. And because the current state of this project lacks proper unit/integration tests I find it very difficult to check if the code still behaves the same after merging your changes. So therefore I suggest that you create several smaller PRs to decrease the size to make it easier for me to notice if some part of the refactor breaks the current behavior. Thanks!

PR #180, opened 2019-08-27
Bugfix/focus not returned

sgruetter:
Fixes an issue where the the previous window's focus was not restored after work in ueli is done.

Reproduce the error without this change merged:

1. Focus on a text editor and type something.
2. Open ueli with the configured shortcut, and do something (e.g. `2+2`).
3. Close ueli by hitting Esc.
4. **Expected:** the text editor has focus restored and the user would without further interaction being able to continue typing. **Actual on Mac:** the text editor window is active, but has no focus. **Actual on Windows:** no window has focus.

Depending on the platform, the focus has returned differently as described at [SO](https://stackoverflow.com/a/55104179). This change implements the expected behavior so that focus is restored correctly.

Tested on current macOS (10.14.6) and Windows 10 (1903) using `yarn bundle` and `yarn start` (no production build).


oliverschwendener:
@sgruetter Thanks for your contribution! :) Looks good to me! But can you target the `dev` branch instead of `master` so this fix can be included in the next release?

oliverschwendener:
Ok nevermind, I change the base to `dev` now.

oliverschwendener:
Hat to revert the changes. `app.hide()` causes all windows to be hidden, the settings window included. So you could not just bring up the main window while having the settings window open.

sgruetter:
Sorry, didn't see your messages as the mails went to an account I'm not checking right now.. [`app.hide`](https://electronjs.org/docs/api/app#apphide-macos) does indeed hide _all_ windows of an application. 

What about invoking `app.hide` only when the settings frame isn't shown? If the settings window is shown, the focus remains with the same application so there should be no noticeable issues from a UX perspective.

If that sounds good to you, I'll give it a try.

oliverschwendener:
Sounds good to me. Also check if your fix will work with the option "Hide window when focus is lost
" disabled.

oliverschwendener:
@sgruetter I opened a new issue for this: #182

Issue #182, opened 2019-08-28
Bug: focus not returned on Windows


oliverschwendener:
Original report from @sgruetter:

Reproduce the error:

- Focus on a text editor and type something.
- Open ueli with the configured shortcut, and do something (e.g. 2+2).
- Close ueli by hitting Esc.

**Expected:** the text editor has focus restored and the user would without further interaction being able to continue typing.

~~**Actual** (on macOS): the text editor window is active, but has no focus.~~ (Fixed)
**Actual** (on Windows): no window has focus.

oliverschwendener:
This is fixed now in [v8.4.0](https://github.com/oliverschwendener/ueli/releases/v8.4.0) (only on macOS). I have to look into this how I can solve this for Windows.

PR #183, opened 2019-08-28
Bugfix/182 focus not returned

sgruetter:
This pull request is an improvement over [PR 180](https://github.com/oliverschwendener/ueli/pull/180) where there were issues regarding the settings window. Thus is was not merged.

On macOS, where `app.hide()` is necessary to return focus to the previous window, a check is made before invocation to skip hiding the app if the settings window is open. As then only the main window is hidden, focus remains with the same application (ueli) and falls back to a previously focused element, if any.

I refactored out the checks if the windows are open to separate functions to easier see what's going on. Also, I added a small comment regarding the special behavior.

Tested on macOS and Windows 10, with settings open and closed, and "Hide window when focus is lost" both on and off.

Also, the source and destination of the PR are `dev`.

@oliverschwendener Now I wonder if you find further issues :-)

Fixes: #182

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -111,7 +111,7 @@ function registerGlobalKeyboardShortcut(toggleAction: () => void, newHotKey: Glo
 }
 
 function showMainWindow() {
-    if (mainWindow && !mainWindow.isDestroyed()) {
+    if (isMainWindowOpen()) {
         if (mainWindow.isVisible()) {
             mainWindow.focus();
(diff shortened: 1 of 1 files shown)


oliverschwendener:
Closing this because I added my own implementation of this in [v8.4.0](https://github.com/oliverschwendener/ueli/releases/v8.4.0). But thanks anyway @sgruetter!
The fix works only on macOS though.

PR #222, opened 2019-11-18
add russian translate

Lukentui:
with Russia from love :3


oliverschwendener:
Thanks for your contribution! 🙌 

There are 2 things missing:
- [ ] Please add Russian to `language.ts`
- [ ] Please add the russian translation set to the switch statement in `translation-set-manager.ts`

Lukentui:
Added)

oliverschwendener:
Can you please make just one PR?

Lukentui:
I'll try

Lukentui:
that's right?

src/common/translation/language.ts
```diff
@@ -1,4 +1,5 @@
 export enum Language {
     English = "English",
     Deutsch = "Deutsch",
+    Russian = "Russian",
```
oliverschwendener: @Lukentui Can you add the russian word for "Russian" between the double quotes?

Lukentui:
like this?

PR #223, opened 2019-11-18
add russian to translation-set-manager.ts

Lukentui:


oliverschwendener:
Closing this because these changes should be in #222

PR #225, opened 2019-11-18
Added option to hide window after execution

oliverschwendener:
Added the user option to control whether the window should stay opened after execution or not.


src/common/translation/russian-translation-set.ts
```diff
@@ -66,6 +66,7 @@ export const russianTranslationSet: TranslationSet = {
     generalSettingsResetWarning: "Вы хотите вернуть все основные настройки по умолчанию. Вы уверены?",
     generalSettingsResetAllSettingsWarning: "Вы хотите вернуть ВСЕ настройки по умолчанию. Вы уверены?",
     generalSettingsClearExecutionLogWarning: "Вы хотите очистить лог выполнения. Вы уверены?",
+    generalSettingsHideMainWindowAfterExecution: "Hide window after execution",
```
oliverschwendener: @Lukentui can you add the correct translation here?

Lukentui: Yesh. Check PR.

Issue #49, opened 2018-08-15
Calculator Plugin to use with , instead of .


thebugfixnet:
Maybe I am blind. But is there no option to switch from . to , in the calculator plugin? It is really annoying for a nearly perfect tool to  use the calculator plugin with the german keyboard layout and , as standard separator for digits.

102.4+102.2 works
102,2+102,2 do not work

oliverschwendener:
Thank you for your feedback!

The problem here is that ueli is using [mathjs](https://github.com/josdejong/mathjs) to evaluate mathematical expressions. To my knowledge mathjs does not support any customization so that a `,` is recognized as a decimal separator. But I will look into this problem as soon as possible.

khanhas:
mathjs also support array and `,` is used to separate array members. I don't think there is anything we can do about that except using another math evaluate package.

oliverschwendener:
@khanhas Thank you for your research!

So apparently there is no way to get mathjs to work with `,` as a decimal separator and I am not willing to change the math library because of this issue. I am sorry but you have to stick with `.` as e decimal separator.

tkohlmeier:
Maybe we could integrate another math parser that supports locales? After a quick search I found this:
https://www.npmjs.com/package/math-and-unit-parser
It supports setting delimiters based on locale or manually. We could add an option in the plugin settings to switch the parser and set delimiters.
What do you think?

oliverschwendener:
Now available in the latest release: https://github.com/oliverschwendener/ueli/releases/tag/v8.10.0

thebugfixnet:
Thank you so much! This is really helpful!!!

I'm really glad you build it in in this way!

PR #232, opened 2019-11-20
Comma as separator in calculator

Lukentui:
#49 
Maybe u talk about another problems, but I don't saw anything in the implementation.
https://imgur.com/a/ah6eoiQ

### DIFF
--- a/src/main/plugins/calculator-plugin/calculator-plugin.ts
+++ b/src/main/plugins/calculator-plugin/calculator-plugin.ts
@@ -56,6 +56,13 @@ export class CalculatorPlugin implements ExecutionPlugin {
     }
 
     private getInputNumber(userInput: string): string {
-        return userInput.replace(",", ".");
+        const inputWithoutCommas: string = userInput.replace(/,/g, ".");
+
+        if (inputWithoutCommas.indexOf("(") === -1 && inputWithoutCommas.indexOf(")") === -1) {
+            return inputWithoutCommas;
+        } else {
+            // replace all `.` to `,` in brackets
+            return inputWithoutCommas.replace(/(\(.*?)\.(.*?\))/g, "$1,$2");
+        }
     }
 }
--- a/src/main/plugins/calculator-plugin/calculator-plugin.ts
+++ b/src/main/plugins/calculator-plugin/calculator-plugin.ts
@@ -20,12 +20,13 @@ export class CalculatorPlugin implements ExecutionPlugin {
     }
 
     public isValidUserInput(userInput: string, fallback?: boolean | undefined): boolean {
-        return Calculator.isValidInput(userInput);
+        return Calculator.isValidInput(this.getInputNumber(userInput));
     }
 
     public getSearchResults(userInput: string, fallback?: boolean | undefined): Promise<SearchResultItem[]> {
         return new Promise((resolve, reject) => {
-            const result = Calculator.calculate(userInput, Number(this.config.precision));
+            const result = Calculator.calculate(this.getInputNumber(userInput), Number(this.config.precision));
+
             resolve([{
                 description: this.translationSet.calculatorCopyToClipboard,
                 executionArgument: result,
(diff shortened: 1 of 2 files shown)


oliverschwendener:
What happens if you do this: `pow(2,6)`?

image

oliverschwendener:
By the way you can add images directly in the GitHub comment ;)

Lukentui:
> 
> 
> By the way you can add images directly in the GitHub comment ;)

It's really conveniently :D

src/main/plugins/calculator-plugin/calculator-plugin.ts
```diff
@@ -56,6 +56,13 @@ export class CalculatorPlugin implements ExecutionPlugin {
...
 
     private getInputNumber(userInput: string): string {
-        return userInput.replace(",", ".");
+        const inputWithoutCommas: string = userInput.replace(/,/g, ".");
+
+        if (inputWithoutCommas.indexOf("(") === -1 && inputWithoutCommas.indexOf(")") === -1) {
+            return inputWithoutCommas;
+        } else {
+            // replace all `.` to `,` in brackets
+            return inputWithoutCommas.replace(/(\(.*?)\.(.*?\))/g, "$1,$2");
```
Lukentui: This solves the problem, but needs testings.

test

Lukentui:
This solves the problem, but needs testings.
img

oliverschwendener:
The calculator plugin heavily depends on [mathjs](https://mathjs.org/) and if we start tweaking the input that the user provides we'll maybe break it's functionality. As far as I know mathjs uses a `,` for separating values in an array evaluation and `.` for the decimal separator.

This is why I'm not willing to merge this PR just for the sake of being able to write `,` instead of `.`.

MetaMmodern:
> What happens if you do this: `pow(2,6)`?
> 
> image

Wow. Hi Oliver and thank you for your great application, I use it as a fast calculator every day))). I was looking for another issue actually. I'd like ueli to automatically detect comma or period delimeter for numbers. I mean not the delimeter for thousands, but for the part of the number which is less than one(like for you it is after period: .23, for me it is after comma: ,23)

But now I see why it is almost impossible, you've implemented functions inside ueli, so commas are parsed as delimeters for expressions, not for numbers. But I have a question anyway. Is there any way you could do this? It is very frustrating to change language every time I open and close ueli. I write mostly in Russian or Ukrainian, but when I open ueli the standard delimeter(comma for me) does not work(

Yes, I know that I can change the delimeter in settings, but consider the opposite situation, when I type smth in a doc in English, open ueli and start fast calculation with a comma, it will be the same, I'd have to change the language again((

(Also, I use numpad when calucaltion something, so the comma button on it changes to point when a switch to English(telling this just in case if it matters))

tkohlmeier:
When the decimal separator is set to comma, the argument separator is semicolon.

So this should work: `pow(2;6)`

MetaMmodern:
> When the decimal separator is set to comma, the argument separator is semicolon.
> 
> So this should work: `pow(2;6)`

Yes, I understand, but consider the opposite situation. I use 2 languages: Russian and English and the seperator on numpad corresponds to different symbols on different languages. In Russian it is comma, but in English it is period. Maybe you have some ideas to use as a workaround for this?

MetaMmodern:
> Alright, I misunderstood you.
> 
> I think ueli should just use the system decimal separator. So instead of having to manually adjust it, it should detect that the input language has changed and change the decimal separator accordingly. That should fit you needs, right?

Yeeeees, totally, right in the spot! That would be great. 
Wow, you've described the solution so clearly, amazing))) 
Are you a contributor to this repo?

tkohlmeier:
Yes, among other things I made the decimal separator selectable (#541).

When I find some time I will try to implement this.

MetaMmodern:
> Yes, among other things I made the decimal separator selectable (#541).
> 
> When I find some time I will try to implement this.

Awesome, mention me in PR then please, I'll request some donation addresses 😄🤡
I write in ts too, but I have no experience in electron and vue, otherwise I'd try to implement it in a fork 😅

Issue #242, opened 2019-12-02
Move core ueli functionality from main to renderer process


oliverschwendener:
Currently ueli uses electron's main process to handle all the core logic. This is not optimal because the main process should only manage the browser windows and can block the UI if it is stuck. The objective of this task is to move all "core" functionality from then main process to the renderer process so the UI won't freeze. This should result in a better user experience and should remove a lot of communication that is currently handled over the ipc module.

oliverschwendener:
This is no longer relevant as the next version of ueli will have `nodeIntegration` truned off in all browser windows due to security reasons and will communicate via [contextBridge](https://www.electronjs.org/docs/api/context-bridge) to the main process, which will handle all the core logic.

PR #250, opened 2019-12-15
created control panel plugin

tkohlmeier:
This is a plugin for the classic Windows control panel.

Some remarks:

- I use Powershell to retrieve all control panel items using _executeCommandWithOutput_, because with node-powershell no result was returned (neither positive nor negative).
- Calling _Get-ControlPanelItem_ takes some time (about 2 seconds on my system), this extends the time needed for refreshing the index.
- Not all control panel item icons can be retrieved form the registry (I don't know where this information is stored...). In that case, the default icon is used.
- I intentionally don't await the powershell command to open a control panel item so that the ueli window is hidden right after pressing enter. Therefore I log errors that occur during execution right in the plugin.
- I created the Russian translations with deepl.com, so they might be not perfect.


src/renderer/settings/plugin-settings.ts
```diff
@@ -16,4 +16,5 @@ export enum PluginSettings {
     ColorConverter = "Color Converter",
     BrowserBookmarks = "Browser Bookmarks",
     Dictionary = "Dictionary",
+    ControlPanel = "Control Panel",
```
oliverschwendener: Please remove your Plugin from `src\renderer\settings\plugin-settings.ts` and add it to `src\renderer\settings\settings-os-specific.ts` with a `win:` prefix. This makes your plugin settings only visible on Windows.

src/main/plugins/control-panel-plugin/control-panel-plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+import { ControlPanelOptions } from "../../../common/config/control-panel-options";
+import { ControlPanelItem } from "./control-panel-item";
+import { defaultControlPanelIcon } from "../../../common/icon/default-icons";
+import { ControlPanelItemsRetriever } from "./control-panel-items-retriever";
+import { Logger } from "../../../common/logger/logger";
+import * as Powershell from "node-powershell";
+
+export class ControlPanelPlugin implements SearchPlugin {
+    public pluginType = PluginType.ControlPanel;
+    private controlPanelItems: ControlPanelItem[] = [];
```
oliverschwendener: Please initialize this empty array in the constructor.

src/main/plugins/control-panel-plugin/control-panel-plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    constructor(private config: ControlPanelOptions, private readonly logger: Logger) {}
```
oliverschwendener: Can you please use the same pattern as the other plugins? Declare class properties before the constructor and then assign the parameters in the constructor. I suggest this:

``` typescript
private controlPanelItems: ControlPanelItem[];
private config: ControlPanelOptions;
private readonly logger: Logger;

constructor(config: ControlPanelOptions, logger: Logger) {
    this.config = config;
    this.logger = logger;
    this.controlPanelItems = [];
}
```

I know that your solution looks cleaner and is less lines of code but I want the code of ueli to be/feel consistent.

src/renderer/settings/plugin-settings.ts
tkohlmeier: Do you mean `win32:` as prefix? The other Windows-only plugins use this, too.

oliverschwendener: Yes I mean `win32:`, sorry my bad

oliverschwendener:
@tkohlmeier Thank you for your contribution! It's very nice to see people creating new plugins for ueli :)
I left a few comments on your changes/additions. Let's have a chat here!

src/main/plugins/control-panel-plugin/control-panel-plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    public async execute(searchResultItem: SearchResultItem, privileged: boolean): Promise<void> {
+        const shell = new Powershell({});
+        shell.addCommand(`powershell -Command "Show-ControlPanelItem -Name '${searchResultItem.executionArgument}'"`);
+        shell.invoke()
+            .catch((reason) => this.logger.error("Opening control panel item failed: " + reason))
+            .finally(() => shell.dispose());
+    }
+
+    public async updateConfig(updatedConfig: UserConfigOptions, translationSet: TranslationSet): Promise<void> {
```
oliverschwendener: I would also like to see the pattern that is used in the other parts of ueli.

Instead of this:

``` typescript
public async updateConfig(updatedConfig: UserConfigOptions, translationSet: TranslationSet): Promise<void> {
    this.config = updatedConfig.controlPanelOptions;
}
```

I would rather have this:

``` typescript
public updateConfig(updatedConfig: UserConfigOptions, translationSet: TranslationSet): Promise<void> {
    return new Promise((resolve) => {
        this.config = updatedConfig.controlPanelOptions;
        resolve();
    });
}
```

Also here, I see that your solution is more elegant and uses the more modern `async` keyword. But I would rather have code/pattern consistency throughout all components.

tkohlmeier:
Thanks for reviewing, I already fixed the issues you found.

What do you think about the time needed for refreshing the control panel index? I once noticed that it can be annoying if you just want to open a program and then have to wait some seconds... Maybe the index refreshing process could be made non-blocking? At least the periodic background refresh. If called using the "Refresh indexes" command it's probably nice to see when it's done.

oliverschwendener:
Yes the index refresh should be done non-blocking. The problem is that most of ueli's functionality is handled in electron's main process. When I started with this project I wasn't aware that you should do these things idealy in the renderer process. So this is a bigger task I have to tackle (see #242).

Anyway: I would make the plugin disabled by default because:

- The control panel is getting more and more a "legacy" feature in Windows
- I suppose more than 50% of the users won't use this feature.
- The normal windows settings plugin is also turned off by default
- Makes the index refresh noticably slower

tkohlmeier:
That's ok with me.

I removed all asyc/awaits now and disabled refreshing index if the plugin is disabled (it would be nice, if those things were handled somewhere else).

oliverschwendener:
The `refreshIndex` method is only called for all activated plugins. See [search-engine.ts#132](https://github.com/oliverschwendener/ueli/blob/1a2878baeb0c4ffa03bf54610303c7ffaecd0eee/src/main/search-engine.ts#L132)

src/main/plugins/control-panel-plugin/control-panel-plugin.ts
```diff
@@ -0,0 +1,91 @@
...
+                supportsAutocompletion: true,
+                supportsOpenLocation: false,
+            }));
+            resolve(searchResultItems);
+        });
+    }
+
+    public refreshIndex(): Promise<void> {
+        return new Promise((resolve, reject) => {
+            if (this.config.isEnabled) {
```
oliverschwendener: This check is unnecessary because the refresh index is called from `search-engine.ts` only if the plugin is enabled. See here: https://github.com/oliverschwendener/ueli/blob/1a2878baeb0c4ffa03bf54610303c7ffaecd0eee/src/main/search-engine.ts#L131-L139

src/main/plugins/control-panel-plugin/control-panel-plugin.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public execute(searchResultItem: SearchResultItem, privileged: boolean): Promise<void> {
+        return new Promise((resolve) => {
+            const shell = new Powershell({});
+            shell.addCommand(`powershell -Command "Show-ControlPanelItem -Name '${searchResultItem.executionArgument}'"`)
+                .then(() => shell.invoke())
+                .catch((reason) => this.logger.error("Opening control panel item failed: " + reason))
+                .finally(() => shell.dispose());
+            resolve();
```
oliverschwendener: I think you should resolve the promise only when the execution was successful. And on error you should reject the promise.

tkohlmeier: I wrote about this in the initial comment:

> I intentionally don't await the powershell command to open a control panel item so that the ueli window is hidden right after pressing enter. Therefore I log errors that occur during execution right in the plugin.

I'm going to change this as you suggested because I think all the plugins should behave the same. But I don't like that the ueli window stays visible that long without any useful feedback if the exeuction takes some time. It feels like something didn't work... Maybe some execution indicator should be shown at least?

src/main/plugins/control-panel-plugin/control-panel-plugin.ts
tkohlmeier: I saw the same check in the application search plugin, that's why I added it...

src/main/plugins/control-panel-plugin/control-panel-plugin.ts
oliverschwendener: Yes you're right, there should be an indicator that "something" is going on when pressing enter. But that's a separate task. Nevertheless, the promise here should only resolve on success and reject on error.

src/main/plugins/control-panel-plugin/control-panel-plugin.ts
oliverschwendener: Alright, that's a mistake then. I'll remove it in the application search plugin.

src/common/config/control-panel-options.ts
```diff
@@ -0,0 +1,7 @@
+export interface ControlPanelOptions {
+    isEnabled: boolean;
+}
+
+export const defaultControlPanelOptions: ControlPanelOptions = {
+    isEnabled: true,
```
oliverschwendener: Can you disable the feature by default?

src/renderer/settings/control-panel-settings-component.ts
```diff
@@ -0,0 +1,71 @@
...
+                    this.updateConfig();
+                },
+                message: translations.resetPluginSettingsToDefaultWarning,
+                modalTitle: translations.resetToDefault,
+                type: UserConfirmationDialogType.Default,
+            };
+            vueEventDispatcher.$emit(VueEventChannels.settingsConfirmation, userConfirmationDialogParams);
+        },
+        updateConfig() {
+            vueEventDispatcher.$emit(VueEventChannels.configUpdated, this.config);
```
oliverschwendener: When activating the plugin the `refreshIndex` method of the plugin should be called but this doesn't happen yet. You should add here a third parameter `true` and a fourth parameter `PluginType.ControlPanel`, like this:

``` typescript
updateConfig() {
    vueEventDispatcher.$emit(VueEventChannels.configUpdated, this.config, true, PluginType.ControlPanel);
}
```

The third parameter specifies if a index refresh is needed or not. The fourth parameter specifies the plugin which should be refreshed.

See here: https://github.com/oliverschwendener/ueli/blob/1a2878baeb0c4ffa03bf54610303c7ffaecd0eee/src/renderer/renderer.ts#L142-L150

oliverschwendener:
@tkohlmeier Thanks alot for your efforts! Looks good to me now. Merging.

Issue #239, opened 2019-12-01
Dictionary definitions missing


stefnotch:
Some dictionary definitions are missing. When searching for `cat`, [the API returns the following json.](https://googledictionaryapi.eu-gb.mybluemix.net/?define=cat&lang=en)
```json
[
    {
        "word": "cat",
        "phonetic": "/kat/",
        "origin": "Old English catt, catte, of Germanic origin; related to Dutch kat and German Katze; reinforced in Middle English by forms from late Latin cattus.",
        "meaning": {
            "noun": [
                {
                    "definition": "A small domesticated carnivorous mammal with soft fur, a short snout, and retractable claws. It is widely kept as a pet or for catching mice, and many breeds have been developed.",
                    "synonyms": [
                        "feline"
                    ]
                },
                {
                    "definition": "(especially among jazz enthusiasts) a man.",
                    "example": "this West Coast cat had managed him since the early 80s"
                },
                {
                    "definition": "A short tapered stick used in the game of tipcat."
                }
            ],
            "verb": [
                {
                    "definition": "Raise (an anchor) from the surface of the water to the cathead.",
                    "example": "I kept her off the wind and sailing free until I had the anchor catted"
                }
            ]
        }
    },
    {
        "word": "cat",
        "phonetic": "/kat/",
        "meaning": {}
    },
    {
        "word": "cat",
        "phonetic": "/kat/",
        "meaning": {}
    },
    {
        "word": "CAT",
        "phonetic": "/kat/",
        "meaning": {
            "abbreviation": [
                {
                    "definition": "short for \"computerized axial tomography\""
                },
                {
                    "definition": "Computer-assisted (or -aided) testing."
                },
                {
                    "definition": "Clear air turbulence."
                }
            ]
        }
    }
]
```

However, Ueli outputs only two definitions.
![image](https://user-images.githubusercontent.com/10220080/69918019-46cc3400-146d-11ea-8c66-debf5d2d798a.png)
 Those two definitions are also the only definitions with an example.

oliverschwendener:
A fix for this will be included in the next release.

Issue #185, opened 2019-08-30
Search Window Closes Only After Delay


sgruetter:
When opening an application, I type the name, see quickly the correct search result, and then hit Enter.

Now 

- the search field gets emptied and, after probably 500 ms,
- the search window gets hidden.

This delay until the the search disappears is a bit annoying. I know, not really a big issue 😄

See in the video: when "sla" disappears, I just hit Enter. A bit later, the entire window gets hidden.

![openafterdelay](https://user-images.githubusercontent.com/30778657/63992330-99532880-caeb-11e9-99a1-d73b6bd95292.gif)

Is there a reason for the delay? Otherwise I offer to look into this, if it's fine with you, @oliverschwendener.

oliverschwendener:
Strangely this seems to appear only on macOS and only after the app is being packaged. When I run the app in development mode, then this delay does not appear.

cosimochellini:
I noticed this problem also on Windows 10

natyusha:
This seems to happen for me when opening most programs on Windows 10 as well.

PR #277, opened 2020-01-27
fix dictionary plugin missing definitions and fixed window hide delay

AmrBinBashir:
1- This fixes the bug of missing definitions or no definitions at all. see #239
the bug was when you filtered the response from the api to only include definitions that have examples but a lot of definitions has no examples. so by removing `&& entry.example` the bug is fixed but now the filter function is useless so i removed it.

2- This fixes the bug of window hide delayed  see #185 and that was because the logic responsible for hiding the window after execution was called in the promise resolve after finishing executing the command in the background. so i moved the logic outside of the promise.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "email": "oliver.schwendener@hotmail.com",
     "url": "https://github.com/oliverschwendener"
   },
-  "description": "This is a keystroke launcher for Windows and macOS.",
+  "description": "A keystroke launcher for Windows and macOS",
   "os": [
     "win32",
     "darwin"
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "type": "git",
     "url": "https://github.com/oliverschwendener/ueli"
   },
-  "homepage": "https://ueli.oliverschwendener.ch",
+  "homepage": "https://ueli.app",
   "author": {
     "name": "Oliver Schwendener",
     "email": "oliver.schwendener@hotmail.com",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "email": "oliver.schwendener@hotmail.com",
     "url": "https://github.com/oliverschwendener"
   },
-  "description": "A keystroke launcher for Windows and macOS",
+  "description": "This is a keystroke launcher for Windows and macOS.",
   "os": [
     "win32",
     "darwin"
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "type": "git",
     "url": "https://github.com/oliverschwendener/ueli"
   },
-  "homepage": "https://ueli.app",
+  "homepage": "https://ueli.oliverschwendener.ch",
   "author": {
     "name": "Oliver Schwendener",
     "email": "oliver.schwendener@hotmail.com",
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -609,7 +609,6 @@ function registerAllIpcListeners() {
         searchEngine.execute(searchResultItem, privileged)
             .then(() => {
                 userInputHistoryManager.addItem(userInput);
-                
             })
             .catch((err) => logger.error(err));
         if (searchResultItem.hideMainWindowAfterExecution && config.generalOptions.hideMainWindowAfterExecution) {
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -609,13 +609,14 @@ function registerAllIpcListeners() {
         searchEngine.execute(searchResultItem, privileged)
             .then(() => {
                 userInputHistoryManager.addItem(userInput);
-                if (searchResultItem.hideMainWindowAfterExecution && config.generalOptions.hideMainWindowAfterExecution) {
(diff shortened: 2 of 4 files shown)


oliverschwendener:
Please target the `dev` branch.

AmrBinBashir:
Oh i thought i did, my bad. fixed it.

package.json
```diff
@@ -7,13 +7,13 @@
...
   "author": {
     "name": "Oliver Schwendener",
     "email": "oliver.schwendener@hotmail.com",
     "url": "https://github.com/oliverschwendener"
   },
-  "description": "A keystroke launcher for Windows and macOS",
+  "description": "This is a keystroke launcher for Windows and macOS.",
   "os": [
     "win32",
     "darwin"
```
oliverschwendener: Your branch is based on `master`, please change that to `dev` too. This is why you get these changes here.

src/main/plugins/dictionary-plugin/dictionary-plugin.ts
```diff
@@ -74,10 +74,6 @@ export class DictionaryPlugin implements ExecutionPlugin {
             const keys = Object.keys(definition.meaning);
             keys.forEach((key) => {
                 definition.meaning[key]
-                    .filter((entry: any) => {
-                        return entry.definition
-                            && entry.example;
-                    })
                     .forEach((entry: any) => {
                         dictionaryResults.push({
                             definition: entry.definition,
```
oliverschwendener: I agree on removing the check for the example, because we don't need it anyway. So I suggest you also remove the example from the interface here:

https://github.com/oliverschwendener/ueli/blob/fe882ca8803896f1133146cdb58f40367b281c06/src/main/plugins/dictionary-plugin/dictionary-plugin.ts#L12-L17

and here:

https://github.com/oliverschwendener/ueli/blob/fe882ca8803896f1133146cdb58f40367b281c06/src/main/plugins/dictionary-plugin/dictionary-plugin.ts#L82-L87

But why did you remove the check for the definition? If we removed this filter function totally we would get dictionary results with empty definitions, which are required here:

https://github.com/oliverschwendener/ueli/blob/fe882ca8803896f1133146cdb58f40367b281c06/src/main/plugins/dictionary-plugin/dictionary-plugin.ts#L92-L102

src/main/main.ts
```diff
@@ -609,13 +609,13 @@ function registerAllIpcListeners() {
...
-                } else {
-                    updateMainWindowSize(0, config.appearanceOptions);
-                }
             })
             .catch((err) => logger.error(err));
+        if (searchResultItem.hideMainWindowAfterExecution && config.generalOptions.hideMainWindowAfterExecution) {
+            hideMainWindow();
+        } else {
+            updateMainWindowSize(0, config.appearanceOptions);
+        }
```
oliverschwendener: I am not sure if this is right. We just hide the window before we even know if the execution was successful or not. I agree that there is not enough visualization on the UI when the execution is in progress, so the user doesn't know what's happening. What are your thoughts on this?

src/main/plugins/dictionary-plugin/dictionary-plugin.ts
AmrBinBashir: Yeah you're right, It never occurred to me that the api would send empty definition field, so it safe to keep the filter for definition only.

src/main/main.ts
AmrBinBashir: well I think its better to hide the window right away, then if the execution wasn't successful we can either show ueli again with a result that tells the user what went wrong or create a new window in electron to do that.

package.json
AmrBinBashir: I reverted these changes, I'm not sure if this is the right way to do because i don't know how to change the pull request to get commits from the dev branch on my fork.

stefnotch:
@oliverschwendener I'd recommend [changing the default branch](https://help.github.com/en/github/administering-a-repository/setting-the-default-branch) to `dev` to prevent this problem from happening. Saves a lot of time and trouble.

AmrBinBashir:
I can close this PR and oliver can make these commits by himself if he approves it. or should i close this PR and open another PR from my dev branch?

stefnotch:
I think closing this PR and making a new one would now be the most elegant approach.

oliverschwendener:
@stefnotch No, I won't change the default branch to dev because master is and should be default. I should rather point out in the [development](https://github.com/oliverschwendener/ueli#development) docs that work should be done in the `dev` branch and create a [PR template](https://help.github.com/en/github/building-a-strong-community/creating-a-pull-request-template-for-your-repository).

@AmrBinBashir You can also use `git reset --soft` to undo your commits but keep your work. Then create a new feature branch out of `dev` locally and commit your changes on the new branch. Push the new branch to your fork and then create a new PR.

AmrBinBashir:
@oliverschwendener I didn't do much anyway on my fork so i deleted it and reforked your repo and pushed my commits to dev and made a new PR

PR #282, opened 2020-02-04
fix dictionary missing definitions

AmrBinBashir:


src/main/main.ts
```diff
@@ -609,13 +609,13 @@ function registerAllIpcListeners() {
...
-                } else {
-                    updateMainWindowSize(0, config.appearanceOptions);
-                }
             })
             .catch((err) => logger.error(err));
+        if (searchResultItem.hideMainWindowAfterExecution && config.generalOptions.hideMainWindowAfterExecution) {
+            hideMainWindow();
+        } else {
+            updateMainWindowSize(0, config.appearanceOptions);
+        }
```
oliverschwendener: I'm still not sure yet if this is the right way to handle the "delay". To me this seems more like a workaround.

I added a change in the `dev` branch which might fix this "delay" issue properly. Instead of using commandline tools to open apps/files/folders ueli will now use electron's [shell](https://www.electronjs.org/docs/api/shell) API which should be much faster. I suggest you revert these changes (regarding the delay) and we observe if my changes fix the issue. If this won't be the case we can look into your solution again.

src/main/plugins/dictionary-plugin/dictionary-plugin.ts
```diff
@@ -75,13 +74,11 @@ export class DictionaryPlugin implements ExecutionPlugin {
             keys.forEach((key) => {
                 definition.meaning[key]
                     .filter((entry: any) => {
-                        return entry.definition
-                            && entry.example;
+                        return entry.definition;
```
oliverschwendener: Looks good to me!

src/main/main.ts
AmrBinBashir: That is definitely better and noticeably faster. I will test it out for a while and report back if there is any issues.

PR #283, opened 2020-02-05
Show suggestion for web search plugins

### LINKED ISSUE #265: Show google search suggestions in websearch plugin
Is it possible to have google search suggestions in websearch plugin?

AmrBinBashir:
Closes #265. 
This PR:
1. adds search suggestions to web search plugin(for now i limited it to 8 suggestion cause frankly I think that's enough).
2. adds default values for suggestion url for all default engines.
3. adds input field and preview for suggestions url. ( Translations for Russian and German are translated by google so they might not be the best. )

**Note: users will have to reset the plugin settings to default first.**

<s>This is not perfect actually, i tried to make it work if a multiple search engines has the same prefix it would get suggestion for every webSearchEngine but that comes with a draw back, say i have two search engines, "duckduckgo and google"  with the same prefix but google's priority is higher, now if google's axios call is successful and duckduck go is not, we get suggestions for google only and  that's fine but if it was the other way around and  google fails first, then duckduckgo's respones won't even get handled as the execution goes straight to the `.catch()` block after the first error and never goes back to the `.then()` block.

The problem in my opinion is that users can set the same prefix for multiple webSearchEngines and even plugins, I think it would be better to have some sorta of a prefix validation that doesn't allow the user to set the same prefix for multiple webSearchEngines and also plugins. That would solve the problem.</s>

### DIFF
--- a/src/main/plugins/websearch-plugin/web-search-engine.ts
+++ b/src/main/plugins/websearch-plugin/web-search-engine.ts
@@ -8,4 +8,5 @@ export interface WebSearchEngine {
     priority: number;
     isFallback: boolean;
     encodeSearchTerm: boolean;
+    suggestionUrl: string;
 }
--- a/src/main/plugins/websearch-plugin/websearch-plugin.ts
+++ b/src/main/plugins/websearch-plugin/websearch-plugin.ts
@@ -43,13 +43,13 @@ export class WebSearchPlugin implements ExecutionPlugin {
                     return 0;
                 });
 
-            webSearchEngines.forEach((webSearchEngine, index) => {
+            webSearchEngines.forEach((webSearchEngine) => {
                 searchResults.push({
-                    description: this.buildDescription(webSearchEngines[index], userInput),
-                    executionArgument: this.buildExecutionArgument(webSearchEngines[index], userInput),
+                    description: this.buildDescription(webSearchEngine, userInput),
+                    executionArgument: this.buildExecutionArgument(webSearchEngine, userInput),
                     hideMainWindowAfterExecution: true,
-                    icon: isValidIcon(webSearchEngines[index].icon) ? webSearchEngines[index].icon : defaultWebSearchIcon,
-                    name: userInput.replace(webSearchEngines[index].prefix, ""),
+                    icon: isValidIcon(webSearchEngine.icon) ? webSearchEngine.icon : defaultWebSearchIcon,
+                    name: userInput.replace(webSearchEngine.prefix, ""),
                     originPluginType: this.pluginType,
                     searchable: [],
                 });
@@ -77,7 +77,7 @@ export class WebSearchPlugin implements ExecutionPlugin {
                         const suggestionsResults: SearchResultItem[] = [];
                         const suggestionsData = response.data[1];
 
-                        suggestionsData.some((suggestion: string) => {
+                        suggestionsData.some((suggestion: string,index:number) => {
                             suggestionsResults.push({
                                 description: this.buildDescription(webSearchEngine, suggestion),
                                 executionArgument: this.buildExecutionArgument(webSearchEngine, suggestion),
@@ -88,7 +88,7 @@ export class WebSearchPlugin implements ExecutionPlugin {
                                 searchable: [],
                             })
 
-                            return suggestion === suggestionsData[7];
+                            return index === 7;
                         })
                         resolve(suggestionsResults)
                     }))
--- a/src/main/plugins/websearch-plugin/websearch-plugin.ts
+++ b/src/main/plugins/websearch-plugin/websearch-plugin.ts
@@ -7,18 +7,21 @@ import { WebSearchEngine } from "./web-search-engine";
 import { TranslationSet } from "../../../common/translation/translation-set";
 import { defaultWebSearchIcon } from "../../../common/icon/default-icons";
 import { isValidIcon } from "../../../common/icon/icon-helpers";
-import axios, { AxiosPromise } from "axios";
+import axios from "axios";
+import { Logger } from "../../../common/logger/logger";
 
 export class WebSearchPlugin implements ExecutionPlugin {
     public readonly pluginType = PluginType.WebSearchPlugin;
     private config: WebSearchOptions;
     private translationSet: TranslationSet;
     private readonly urlExecutor: (url: string) => Promise<void>;
+    private readonly logger: Logger;
 
-    constructor(userConfig: WebSearchOptions, translationSet: TranslationSet, urlExecutor: (url: string) => Promise<void>) {
+    constructor(userConfig: WebSearchOptions, translationSet: TranslationSet, logger: Logger, urlExecutor: (url: string) => Promise<void>) {
         this.config = userConfig;
         this.translationSet = translationSet;
         this.urlExecutor = urlExecutor;
+        this.logger = logger;
     }
 
     public getSearchResults(userInput: string, fallback?: boolean): Promise<SearchResultItem[]> {
@@ -39,6 +42,7 @@ export class WebSearchPlugin implements ExecutionPlugin {
                     }
                     return 0;
                 });
+
             webSearchEngines.forEach((webSearchEngine, index) => {
                 searchResults.push({
                     description: this.buildDescription(webSearchEngines[index], userInput),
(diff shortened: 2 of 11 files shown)


AmrBinBashir:
Weird bug with google suggestions:
if i make an axios call from within the search plugin like this 
```
axios
    .get(
        "https://www.google.com/complete/search?client=opera&q=python tutorial"
    )
    .then(res => {
        console.log(res.data[1]);
    })
    .catch(err => {
        console.log(err);
    });
``` 
I get this result 
```
['g. van rossum python tutorial',0]
```
but if i make a `test.js` file and execute it directly with `node test.js` with this content i get this expected result 
```
[  'python tutorial',
  'python tutorial pdf',
  'python tutorial for beginners',
  'python tutorial point',
  'python tutorial شرح',
  'python tutorial guru',
  'python tutorial for beginners pdf',
  'python tutorial point pdf',
  'python tutorial arabic'
]
```
I can't figure out why this happens

oliverschwendener:
I have some questions:
- How does this work with web search engines without suggestions functionality?
- How does this work if the web search engine is set as fallback?

AmrBinBashir:
@oliverschwendener 
1- If there is no URL provided, the plugin wouldn't provide any search results not even the default one cause I had to add the default result after I get responses, that way I can add default results accordingly to every webSearchEngine. This is not perfect and right now I can't think of any better way to do it.

It would make it much easier to have some sort of validation that prevents users from having same prefix for multiple search engines and plugins too and maybe only one fallback search engine.

2- The search engines provide suggestions even when they are fallback.

src/main/plugins/websearch-plugin/websearch-plugin.ts
```diff
@@ -7,6 +7,7 @@ import { WebSearchEngine } from "./web-search-engine";
 import { TranslationSet } from "../../../common/translation/translation-set";
 import { defaultWebSearchIcon } from "../../../common/icon/default-icons";
 import { isValidIcon } from "../../../common/icon/icon-helpers";
+import axios from "axios";
```
oliverschwendener, line 10: I don't want the websearch plugin to depend on axios. Can you somehow abstract this so it can be unit tested in the future?

src/main/plugins/websearch-plugin/websearch-plugin.ts
```diff
@@ -22,7 +23,8 @@ export class WebSearchPlugin implements ExecutionPlugin {
 
     public getSearchResults(userInput: string, fallback?: boolean): Promise<SearchResultItem[]> {
         return new Promise((resolve, reject) => {
-            const searchResults = this.config.webSearchEngines
+            const searchResults: SearchResultItem[] = [];
+            const webSearchEnginesArray = this.config.webSearchEngines
```
oliverschwendener, line 27: In my opinion this variable could be called `webSearchEngines` instead of `webSearchEnginesArray`. This should be self self-explanatory and the IDE can tell you of what type the variable is.

src/main/plugins/websearch-plugin/web-search-engine.ts
```diff
@@ -8,4 +8,5 @@ export interface WebSearchEngine {
     priority: number;
     isFallback: boolean;
     encodeSearchTerm: boolean;
+    suggestionUrl: string;
```
oliverschwendener, line 11: I would mark this as nullable because not every search engine supports suggestions.

src/main/plugins/websearch-plugin/websearch-plugin.ts
```diff
@@ -36,20 +38,46 @@ export class WebSearchPlugin implements ExecutionPlugin {
...
                         hideMainWindowAfterExecution: true,
-                        icon: isValidIcon(webSearchEngine.icon) ? webSearchEngine.icon : defaultWebSearchIcon,
-                        name: webSearchEngine.name,
+                        icon: isValidIcon(webSearchEnginesArray[responseCounter].icon) ? webSearchEnginesArray[responseCounter].icon : defaultWebSearchIcon,
+                        name: userInput.replace(webSearchEnginesArray[responseCounter].prefix, ""),
                         originPluginType: this.pluginType,
                         searchable: [],
-                    };
-                });
+                    });
```
oliverschwendener, line 58: You don't need a variable `responseCounter` here, you get the current counter from the second parameter of the `forEach` function:

`responses.forEach(response, index) => ...`

oliverschwendener: Additionally `res` is a bad variable name, I suggest you rename it to `response` so it is clear what it is.

src/main/plugins/websearch-plugin/websearch-plugin.ts
```diff
@@ -36,20 +38,46 @@ export class WebSearchPlugin implements ExecutionPlugin {
...
+                                executionArgument: this.buildExecutionArgument(webSearchEnginesArray[responseCounter], suggestion),
+                                hideMainWindowAfterExecution: true,
+                                icon: isValidIcon(webSearchEnginesArray[responseCounter].icon) ? webSearchEnginesArray[responseCounter].icon : defaultWebSearchIcon,
+                                name: suggestion,
+                                originPluginType: this.pluginType,
+                                searchable: [],
+                            });
+                        if (results.length > 8) {
+                            return suggestion === results[7];
+                        }
```
oliverschwendener, line 73: Can you explain what is happening here?

src/main/plugins/websearch-plugin/websearch-plugin.ts
AmrBinBashir: I will do that.

src/main/plugins/websearch-plugin/web-search-engine.ts
AmrBinBashir: Okay.

src/main/plugins/websearch-plugin/websearch-plugin.ts
AmrBinBashir: I can use https that comes with node.

src/main/plugins/websearch-plugin/websearch-plugin.ts
AmrBinBashir: When I was writing this code, I was considering that there might be multiple search engines with the same prefix so I limited each search engine to 8 results only (I think this is enough) so user can see other results from other engines. 

The `if` check is a bad code. I should remove it.

src/main/plugins/websearch-plugin/websearch-plugin.ts
AmrBinBashir: I will change it.

src/main/plugins/websearch-plugin/web-search-engine.ts
AmrBinBashir: wouldn't that make the tscompiler complain ? since axios or anyother GET method would expect string .

AmrBinBashir: Wouldn't that make the tscompiler complain? since axios or any GET method would expect a string.

src/main/plugins/websearch-plugin/websearch-plugin.ts
AmrBinBashir: I can use the https module that comes with nodejs but I will have to write a function to promisify its GET method.

src/main/plugins/websearch-plugin/web-search-engine.ts
oliverschwendener: If there is no suggestion url there is no need to fetch suggestions, right?

src/main/plugins/websearch-plugin/websearch-plugin.ts
oliverschwendener: That's not what I mean. When doing actual HTTP requests axios is fine. When this plugin depends on axios or any other http client we can't unit test it's functionality without making actual HTTP requests. This is something I want to avoid. I would rather abstract the component that fetches suggestions, so we're able to mock it when unit testing.

oliverschwendener: For example, take a look at the `urlExecutor` within this class. The plugin itself does not know how the implementation of the execution of URLs is implemented (because it differs depending on which operating system you are). We inject the actual url executor from outside into this class.

See here: https://github.com/oliverschwendener/ueli/blob/5377020897af4992cbe0fe4f1546f9ba5122319c/src/main/plugins/websearch-plugin/websearch-plugin.ts#L11-L21

AmrBinBashir: I see what you mean. But I don't know how can I resolve searchResults outside of axios call.
If ueli has some sorta of an updateResults method that I trigger when axios call finishes and updates the results shown to user.

src/main/plugins/websearch-plugin/web-search-engine.ts
AmrBinBashir: I can resolve empty suggestion for webSearchEngines with no suggestion url.

src/main/plugins/websearch-plugin/websearch-plugin.ts
oliverschwendener: How about something like this?

``` typescript
public getSearchResults(userInput: string, fallback?: boolean): Promise<SearchResultItem[]> {
        const suggestionPromises = this.config.webSearchEngines.map((webSearchEngine) => {
            return this.suggestionResolver(webSearchEngine.suggestionUrl);
        });

        Promise.all(suggestionPromises)
            .then((results) => resolve(results));
            .catch((error) => reject(error))
    });
}
```

AmrBinBashir: I already made this in my last commit except wrapping it in a suggestionResolver function but I resolve the results whether whether `Promise.all` is successful or not, because if one of the suggestionPromises fails the whole plugin would fail. 
So I abstracted the default results outside of Promise.all that way if anything goes wrong I can always show the user the default results, what do you think ? should I keep it this way or should I reject the whole plugin's results if the GET call fails ?

I will make another commit to clean up the code into a suggestionResolver function.

AmrBinBashir: Check my last commit where I cleaned up the code:
suggestionResolver function that returns a promise with suggestionsResults and if a webSearchEngine has no suggestionUrl or the GET request fails, the function returns empty promise. That way the

tmueller:
Thanks for the effort on both sides. Is there anything missing in this PR? Could it be merged?

AmrBinBashir:
Unless oliver has other changes in mind, I think the PR is ready for merge.

oliverschwendener:
Because I was still not happy with the implementation I implemented it here: 59baa7e7f616854906a33879ada3a06c00d91e49

Thanks for your efforts though for finding out how to get suggestions and sorry the nit picking but I'd like to have ueli's code base as clean as possible.

AmrBinBashir:
I understand, the important thing is the feature is implemented the right way.

PR #285, opened 2020-02-06
Fix Catalina missing system apps

underthecocotree:
Prior to this change, while searching to find system apps there were no hits.  Catalina seems to have the applications under `System/Applications`
eg Calendar.app Mail.app

This change adds the `System/Applications` folder to the default application search folders

Issue: oliverschwendener#278

### DIFF
--- a/src/common/config/application-search-options.ts
+++ b/src/common/config/application-search-options.ts
@@ -27,6 +27,7 @@ const macOsApplicationSearchOptions: ApplicationSearchOptions = {
         "/Applications",
         "/System/Library/CoreServices",
         `${homedir()}/Applications`,
+        "/System/Applications",
     ],
     enabled: true,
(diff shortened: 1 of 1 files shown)


underthecocotree:
You mentioned that before and I kind of agree.

Should I add a function in `src/common/helpers/operating-system-helpers.ts` that checks for the version of mac running.

Then based on that add the location to the `applicationFolders` array.

Or we can test if the directory exists and then added that way we don't really have to check for specific os version.

What do you think?

oliverschwendener:
I wouldn't check if the directory exists because that code should be asynchronous and this will complicate the synchronous code that merges the user configuration with the default.

I would use nodejs's `os.release()` API to check the current version of the operating system.

oliverschwendener:
Will close this due to inactivity.

Issue #298, opened 2020-02-18
Dictionary plugin shows incorrect results if i type fast


AmrBinBashir:
when i use the dictionary result, sometimes when i type fast the results is incoreect like if am searching for "bang" it gives me results for "ban". This can be avoided by using a cancel token for axios request so the only request that gets through is the last one thus getting correct results, and maybe this can be  used with other plugins that rely on axios.

PR #299, opened 2020-02-23
add cancelToken and handle its err in dict plugin

AmrBinBashir:
This PR fixes the bug mentioned here #298 by using a CancelToken and ignoring its err in the promise.


AmrBinBashir:
> In general I don't want the dictionary plugin to depend on axios. Is there a more generic way to handle this?

Only other way I know that fixes the bug is using the [debounce](https://www.npmjs.com/package/debounce)package.

PR #300, opened 2020-02-25
Added setVisibleOnAllWorkspaces for MacOS

BenoitGour:
Added setVisibleOnAllWorkspaces for MacOS to prevent workspace switching when toggling ueli.

### DIFF
--- a/src/common/config/general-options.ts
+++ b/src/common/config/general-options.ts
@@ -18,6 +18,7 @@ export interface GeneralOptions {
     rescanIntervalInSeconds: number;
     showAlwaysOnPrimaryDisplay: boolean;
     showTrayIcon: boolean;
+    showOnAllWorkSpaces: boolean;
 }
 
 export const defaultGeneralOptions: GeneralOptions = {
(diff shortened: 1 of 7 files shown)


BenoitGour:
Also added option to toggle the option under General

oliverschwendener:
Thanks for your contribution! I'll look into this and leave some comments.

oliverschwendener:
Is there reason why we shouldn't set `setVisibleOnAllWorkspaces()` always to true? Is there a situation where a user might want the window to appear on another workspace?

BenoitGour:
I don't think there is any reason to not have it always set to true,  which is what I did in my first commit. But i decided to add the option just in case.

oliverschwendener:
Then why don't we just add `mainWindow.setVisibleOnAllWorkspaces(true);` here? https://github.com/oliverschwendener/ueli/blob/ff7dd7727135d969660518e3a132b507bfd3440b/src/main/main.ts#L426-L453

Without any settings and so on?

BenoitGour:
I Removed the options I made. I Added mainWindow.setVisibleOnAllWorkspaces(true, {visibleOnFullScreen: true}); in createMainWindow.
I also included the option visibleOnFullScreen: true   to prevent workspace switching when using fullscreen apps (Parallels Desktop for example)
https://github.com/oliverschwendener/ueli/pull/300/commits/3f195cd3efd9d1c8ac3e8365e79864df68735475

BenoitGour:
fixes issue #303

Issue #318, opened 2020-03-29
Scroll item by item instead of current scrolling mechanism.


AmrBinBashir:
I think it would be better to scroll item by item rather than a whole bunch of them at a time. I can open a pull request with the changes if you want. It's actually pretty simple change.

oliverschwendener:
To be honest I don't agree that this is a better behavior. I will close your PR because I think the current solution is better. If more people support your opinion, we can open the PR later again.

AmrBinBashir:
That sounds reasonable.

dmratcliffe:
I think this would be nice as a setting.
Scroll page vs scroll one by one

PR #320, opened 2020-03-29
add item by item scroll instead of old scroll

AmrBinBashir:
See #318


oliverschwendener:
Closing this PR because I think the current solution is better.

PR #335, opened 2020-04-11
made control panel plugin compatible with Powershell 4.0

tkohlmeier:
This makes the plugin work under Windows 8.1 where by default Powershell 4.0 is installed.


oliverschwendener:
Thanks!

PR #336, opened 2020-04-11
improve control panel item execution feedback

tkohlmeier:
By using child_process.execSync, the "loadingVisible" icon is shown until the execution completes.

Example:
example

### DIFF
--- a/src/main/plugins/control-panel-plugin/control-panel-plugin.ts
+++ b/src/main/plugins/control-panel-plugin/control-panel-plugin.ts
@@ -8,7 +8,7 @@ import { ControlPanelOptions } from "../../../common/config/control-panel-option
 import { ControlPanelItem } from "./control-panel-item";
 import { defaultControlPanelIcon } from "../../../common/icon/default-icons";
 import { ControlPanelItemsRetriever } from "./control-panel-items-retriever";
-import * as Powershell from "node-powershell";
+import * as childProcess from "child_process";
 
 export class ControlPanelPlugin implements SearchPlugin {
     public pluginType = PluginType.ControlPanel;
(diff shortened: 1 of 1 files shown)


src/main/plugins/control-panel-plugin/control-panel-plugin.ts
```diff
@@ -26,12 +26,12 @@ export class ControlPanelPlugin implements SearchPlugin {
...
-                .then(() => shell.invoke())
-                .then(() => resolve())
-                .catch((reason) => reject(reason))
-                .finally(() => shell.dispose());
+            try {
+                childProcess.execSync(`powershell -Command "Show-ControlPanelItem -Name '${searchResultItem.executionArgument}'"`);
+                resolve();
+            } catch (error) {
+                reject(error);
+            }
```
oliverschwendener, line 34: Why using `execSync` instead of `exec`? Won't cause a freeze on the UI?

tkohlmeier: It does not freeze the UI. But it is blocking for the search for the empty string that is performed after pressing enter (because the input text box is cleared). That's why the search spinner is shown until the process execution finishes. It's the same principle as with shell.openItem() (which is blocking as well) in openFile() in file-path-executor.ts.

I know this is hacky and should be replaced by a function that intentially shows the spinner until the execution promise finishes. But because execution of control panel items is quite slow, this is just good for user experience.

oliverschwendener:
I now added a "loader" that shows up until the execution succeeds. I think the problem your PR is adressing would be resolved then. Can you check the latest dev branch?

tkohlmeier:
That looks good! It's even better because the indicator now is shown immediately instead of after the `showLoaderDelay`.

One difference is that the suggestion list is not grayed out while the exeuction spinner is shown. But I think that's ok. I'm closing the PR.

PR #337, opened 2020-04-11
Extended UWP plugin to provide all apps

tkohlmeier:
This extends the UWP plugin to provide all installed apps. It uses the Powershell commands `Get-StartApps`, `Get-AppxPackage` and `Get-AppxPackageManifest` to retrieve the apps including display names and icons.

Extracting the icons was tricky, but it now works for all my apps under Windows 10 and Windows 8.1.


oliverschwendener:
You sir, are a hero! Thanks for this! I'll look at the code and leave some comments/suggestions.

src/main/plugins/uwp-plugin/uwp-apps-retriever.ts
```diff
@@ -0,0 +1,147 @@
...
+import { IconType } from "../../../common/icon/icon-type";
+
+export class UwpAppsRetriever {
+    public static getAll(alreadyKnownApps: UwpApplication[]): Promise<UwpApplication[]> {
+        return new Promise((resolve, reject) => {
+            const alreadyKnownAppIds = alreadyKnownApps.length === 0 ? "@()" : alreadyKnownApps.map((a) => `\\"${a.appId}\\"`).join(",");
+            const command = this.getUwpAppsCommand
+                .replace(/\n/g, " ").replace(/\"/g, "\\\"")
+                .replace("%alreadyKnownAppIds%", alreadyKnownAppIds);
+            this.executeCommandWithUtf8Output(`powershell -Command "${command}"`)
```
oliverschwendener, line 12: I'd like to have a line break here.

src/main/plugins/uwp-plugin/uwp-plugin.ts
```diff
@@ -4,38 +4,40 @@ import { UserConfigOptions } from "../../../common/config/user-config-options";
...
         this.config = config;
-        this.uwpAppRepository = uwpAppRepository;
         this.filePathExecutor = filePathExecutor;
+        this.uwpApps = [];
     }
 
     public getAll(): Promise<SearchResultItem[]> {
+        return new Promise((resolve) => resolve(this.uwpApps.map((app) => this.createSearchResult(app))));
+    }
+    public refreshIndex(): Promise<void> {
```
oliverschwendener, line 28: I'd like to have a line break here.

src/main/plugins/uwp-plugin/uwp-plugin.ts
```diff
@@ -4,38 +4,40 @@ import { UserConfigOptions } from "../../../common/config/user-config-options";
...
-                    resolve(uwpApps.map((uwpApp) => this.createSearchResult(uwpApp)));
+            UwpAppsRetriever
+                .getAll(this.uwpApps)
+                .then((apps) => {
+                    this.uwpApps = apps;
+                    resolve();
                 })
-                .catch((err) => reject(err));
+                .catch((error) => reject(error));
         });
```
oliverschwendener, line 37: I suggest to reformat this code to this (for better readability):

``` typescript
UwpAppsRetriever.getAll(this.uwpApps)
    .then((apps) => {
        this.uwpApps = apps;
        resolve();
    })
    .catch((error) => reject(error));
```

src/main/plugins/uwp-plugin/uwp-plugin.ts
```diff
@@ -4,38 +4,40 @@ import { UserConfigOptions } from "../../../common/config/user-config-options";
...
     }
-    public refreshIndex(): Promise<void> {
-        return new Promise((resolve) => resolve());
-    }
     public clearCache(): Promise<void> {
+        this.uwpApps = [];
```
oliverschwendener, line 40: Don't think this is necessary.

src/main/plugins/uwp-plugin/uwp-plugin.ts
```diff
@@ -4,38 +4,40 @@ import { UserConfigOptions } from "../../../common/config/user-config-options";
...
-                .then((uwpApps) => {
-                    resolve(uwpApps.map((uwpApp) => this.createSearchResult(uwpApp)));
+            UwpAppsRetriever
+                .getAll(this.uwpApps)
+                .then((apps) => {
+                    this.uwpApps = apps;
+                    resolve();
                 })
-                .catch((err) => reject(err));
+                .catch((error) => reject(error));
```
oliverschwendener, line 36: I'd like to have the UwpPlugin not depend on UwpAppsRetriever. How about having the UwpPlugin depend on a function that returns all UWP apps (similar to the `filePathExecutor`). The advantage with this solution would be that the UwpPlugin class could be unit tested much easier in a later stage because it does not depend on the hardcoded UwpAppRetriever class.

Here is an example:

uwp-plugin.ts
``` typescript
class UwpPlugin {
    private uwpAppsRetriever: () => Promise<UwpApplication[]>;

    constructor(uwpAppsRetriever: () => Promise<UwpApplication[]>) {
        this.uwpAppsRetriever = uwpAppsRetriever;
    }

    public getAll(): Promise<SearchResultItem[]> {
        return new Promise((resolve, reject ) => {
            this.uwpAppsRetriever()
                .then((uwpApps) => resolve(this.createSearchResults(uwpApps)))
                .catch((error) => reject(error));
        })
    }

    private createSearchResults(uwpApps: UwpApplication[]): SearchResultItem[] {
        return uwpApps.map((uwpApp): SearchResultItem => {
            return {
                // create search result item here
            };
        })
    }
}
```

uwp-apps-retriever.ts
``` typescript
function convertNewAppToUwpApplication() {
    // stuff from "private static convertNewAppToUwpApplication" goes here
}

function executeCommandWithUtf8Output() {
    // stuff from "private static executeCommandWithUtf8Output" goes here
}

const uwpAppsPowershellCommand = ``; // powershell command goes here

export function getAllUwpApps(): Promise<UwpApplication[]> {
    // get all uwp apps here
}
```

production-search-engine.ts
``` typescript
import { getAllUwpApps } from "./uwp-apps-retriever";

searchPlugins.push(new UwpPlugin(getAllUwpApps));
```

tkohlmeier: You probably meant to call the uwpAppsRetriever in refreshIndex(), didn't you?

oliverschwendener: Sorry, yes :)

tkohlmeier: Alright, that makes sense. I pushed the change.

PR #339, opened 2020-04-13
Calculator: Custom decimal and argument separators

tkohlmeier:
This adds settings for custom decimal and argument separators.

Inspired by this suggestion:
https://mathjs.org/examples/browser/custom_separators.html.html

### DIFF
--- a/src/common/config/calculator-options.ts
+++ b/src/common/config/calculator-options.ts
@@ -1,9 +1,13 @@
 export interface CalculatorOptions {
     precision: number;
+    decimalSeparator: string;
+    argumentSeparator: string;
     isEnabled: boolean;
 }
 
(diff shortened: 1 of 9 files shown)


tkohlmeier:
Any thoughts on this?

tkohlmeier:
I replaced the text inputs with buttons for selection of decimal and argument separators. This prevents the user from entering separators that do not make sense, like digits or math symbols.

oliverschwendener:
In my opinion the decimal separator should also be applied in other plugins where numbers are entered, for example the currency converter.

tkohlmeier:
You are right, this should be a global setting.

What about the argument separator? Since it is calculator-specific., should we remove the setting and automatically set it according to the decimal separator?

tkohlmeier:
Closing this in favor of #541

Issue #314, opened 2020-03-22
Ueli 8.6.2 still not showing the icons of apps on MacOS.


zohozer:
Hi. I just installed the latest version of the Ueli 8.6.2 who it is supposed that fixed the error of not displaying the icons of the apps on MacOS. I just uninstalled the older version using AppCleaner, installed the new version of Ueli 8.6.2 and still I can't see the icons of the app displayed on the left of the names in result field.

<img width="401" alt="Ueli_0 8 2_1" src="https://user-images.githubusercontent.com/5983791/77247682-4974b400-6c2b-11ea-8c08-dc7c834e8b4b.png">

After a clean install:
<img width="604" alt="Ueli_0 8 2_2" src="https://user-images.githubusercontent.com/5983791/77247685-4c6fa480-6c2b-11ea-8f15-3517270543ff.png">

![About_Mac](https://user-images.githubusercontent.com/5983791/77247712-99537b00-6c2b-11ea-999f-9dade547eec3.png)

zohozer:
I just turned the "Use native icons" of and on, and when switching ON, I get this error message:

Error: Command failed: sips -s format png "/Applications/FreeTube.app/Contents/Resources/electron.icns" --out "/Users/winner/.ueli/application-icons/freetube-5ae68a60d8ecf80f1f1301c169ec9a60.png"

oliverschwendener:
Try to clear all caches and then refresh indexes:

![image](https://user-images.githubusercontent.com/15727229/77247995-66fb4b00-6c36-11ea-8d39-316eafa46c72.png)

![image](https://user-images.githubusercontent.com/15727229/77248003-78dcee00-6c36-11ea-8a0a-1ccc974a24a8.png)

zohozer:
I made this a few times, still nothing.

oliverschwendener:
What happens if you uninstall this app temporarily?

zohozer:
I already made this a few times. Still the same. Seems that the icons are generated, as I have a folder with more than 207mb of icons, but for some reason, Ueli is unable to display them? I also found a temp. dir. used by Ueli called "IrisPro", meantime my active GPU is the GT750M. Can this be the issue?

<img width="1217" alt="Ueli_icons" src="https://user-images.githubusercontent.com/5983791/77251306-89489500-6c45-11ea-95ab-786b28ad85c7.png">
[...]

PR #341, opened 2020-04-14
Fix for Issue #314

BenoitGour:
Fixes issue with loading certain icons, if icon loading fails, reverts to defaultIcnsFilePath for generic icon, see issue #314


BenoitGour:
The problem doesn't seem to be with Ueli itself, but with MacOS,  the command sips returns an error even when executed manually,  and if you browse to the .icns location and try to open it with Preview, it gives you an error.

oliverschwendener:
Thanks!

PR #345, opened 2020-04-18
Fix search results scrolling

tkohlmeier:
The removed id of the user input control (commit 5a4fbd7c7d) broke scrolling on arrow-down/up in search results (search-results-component scrollIntoView() uses the id to get the element).


oliverschwendener:
Thanks!

Issue #258, opened 2020-01-04
PageUP and PageDown to switch between pages


zohozer:
On my MacOS 10.14.6 installation of Ueli, seems that it is not possible to use the PageUP and PageDown keys to quickly switch between the page results if there are more results that are filling multiple pages. I can only use the Up and Down arrow keys which is much slower when dealing with a lot of results.

oliverschwendener:
I have no plans to implement this, therefore I'll close this issue.

PR #346, opened 2020-04-18
Allow browsing search results with page up/down and fix scrolling issues

tkohlmeier:
This mainly addresses feature request #258. Search results can now be browsed using page up/down keys.

While working on this, I found some issues with scrolling:

- When you press arrow down till you get to the next page, the scrolling offset isn't correct. There is a little gap above the first search result:

  grafik

- When you are on the second page and go to the second result and then go one back up, ueli unnecessarily scrolls one page up.

These issues should be fixed now, too.

This also makes PR #345 obsolete.


tkohlmeier:
Any thoughts on this?

oliverschwendener:
Sorry for that massive delay, but I finally found some time to look into this 😅 

The page down action does not seem to work properly when I change the option "Max search results per page" in the appearance settings (in my example it was set to `4`). It seems to jump too far, here is a video:

(First I just used the arrow keys, and then I used the page down key)

scrolling

tkohlmeier:
This happened because when the user enters a value for maxSearchResultsPerPage, it is saved as string instead of as number. This resulted in a string concatenation instead of an addition or subtraction.

Fixed this by parsing maxSearchResultsPerPage as number.

But would be nice if maxSearchResultsPerPage was saved as a number...

oliverschwendener:
Ah yes you're right. Because it's an HTML input field it will be parsed into a string 🤦 I'll look into saving the property as a number. Thanks alot for your efforts, your change looks good now!

PR #358, opened 2020-04-26
ignore Windows application searcher errors

tkohlmeier:
The application searcher scans directories defined by the user. These directories can contain files that cannot be accessed and the directories could not exist anymore. Currently this causes the application searcher to throw an error so that no applications are indexed at all. I think we can just ignore all errors.


oliverschwendener:
I'm note sure if this is the right way to handle this. If an application can't be found due to missing permissions, we'll never know that this was the reason why the application is not showing up.

tkohlmeier:
I added logging of errors occurring during application search.

Fooughhy:
There have been several issues regarding this missing functionality (e.g by me) so some kind of valid solution for this would be appreciated!

PR #371, opened 2020-05-07
Disable powershell profile loading for retrievers

### LINKED ISSUE #370: UWP Apps
When I enable UWP Apps I get the following error:
```
SyntaxError: Unexpected token W in JSON at position 0
```

Im using the latest version of **ueli**

pecigonzalo:
Unfortunately, I cant compile this from Linux apparently (or WSL2) and I dont have the tools to do/test on Windows.

Fixes #370

### DIFF
--- a/src/main/executors/file-searchers.ts
+++ b/src/main/executors/file-searchers.ts
@@ -19,7 +19,7 @@ export function windowsFileSearcher(option: FileSearchOption): Promise<string[]>
         const recursive = option.recursive ? "-Recurse" : "";
         const powershellScript = `Get-ChildItem -Path '${option.folderPath}' ${recursive} | Where { ${excludes} } | % { $_.FullName } | Select-Object -First 1000`;
 
-        executeCommandWithOutput(`powershell -Command "& { ${powershellScript} }"`)
+        executeCommandWithOutput(`powershell -NonInteractive -NoProfile -Command "& { ${powershellScript} }"`)
             .then((data) => {
                 const result = data
                     .split("\n")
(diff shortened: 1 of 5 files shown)


oliverschwendener:
Thanks for your contribution! How do you know that this will fix the issue, when you can't compile this? 😄

pecigonzalo:
The error is `SyntaxError: Unexpected token W in JSON at position 0` so its finding the W from the `WARNING: Missing git support, install posh-git with 'Install-Module posh-git' and restart the terminal.` in my console.

While those are not passed on the `|` between `Get-ControlPanelItem | ConvertTo-Json` they are part of the output that ueli tries to parse as JSON.

I then verified that installing `Install-Module posh-git` removes the message and fixes ueli by just installing it and restarting ueli.

With that confirmation, I uninstalled the module again, confiemed the error is back and ran powershell as `powershell -NonInteractive -NoProfile dir` to verify the options disabled the error, which makes sense since the error comes from a module in my profile, and for running ad-hoc command that are generic to every system we should not load the profile.
Ensuring the script is non-interactive and does not load the profiles is common when running background bash scripts or other things in linux and probably here as well.

You can verify this by adding this to your `"$HOME\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1"`
```
Write-Warning "This Warning"
```
This will mimic the warning Im getting and it will break your ueli unless you add my fix or remove the warning.

oliverschwendener:
I see now. Thanks for this!

oliverschwendener:
Can you target my `dev` branch?

pecigonzalo:
Updated

oliverschwendener:
Thanks!

Issue #308, opened 2020-03-06
Error when activating Application Search


pauLee:
I can not launch any application with Ueli and I get an error when activating or modifying the Application search.

The error:
{ "cmd": "mdfind \"kind:apps\" | egrep \"^\\/Applications\"", "code": 1, "killed": false, "level": "error", "service": "user-service", "signal": null, "timestamp": "2020-03-06T11:36:49.611Z" }
<img width="570" alt="Bildschirmfoto 2020-03-06 um 12 38 20" src="https://user-images.githubusercontent.com/592889/76080540-68c2de80-5fa7-11ea-8c67-fb3a9514527b.png">

I'am using Ueli version 8.5.3 on the latest Mac OS 10.15.3
[debug.log](https://github.com/oliverschwendener/ueli/files/4298104/debug.log)

blmhemu:
The bug : You probably might have disabled indexing which this app is using.
https://www.technipages.com/macos-sierra-enable-disable-spotlight

Solution : Try running `sudo launchctl load -w /System/Library/LaunchDaemons/com.apple.metadata.mds.plist`

pauLee:
Thank you for that idea but this wasn't the problem. The spotlight-index isn't disabled. I also tried to rebuild the index. Spotlight itself works very good.

oliverschwendener:
What happens if you reset the application search settings to default?

![image](https://user-images.githubusercontent.com/15727229/77054595-45992580-69d0-11ea-9a29-2bdcd99fde30.png)

pauLee:
The error still remains.

oliverschwendener:
What happens if you remove all folders in the settings?

pauLee:
The error disappears when I delete the last folder path. But it doesn't matter which path I remove last.
And the error does not occur if I disable and enable the plugin when all folders are deleted.

dkanas:
@oliverschwendener I might have found the issue:

I'm running mac OS 10.15.3 and running `mdfind kind:apps` in the terminal returns nothing, but when I run `mdfind apps` it seems to return a full list of apps.

pauLee:
> @oliverschwendener I might have found the issue:
> 
> I'm running mac OS 10.15.3 and running `mdfind kind:apps` or `mdfind "kind:apps"` in the terminal returns nothing, but when I run `mdfind kMDItemContentTypeTree=com.apple.application-bundle` it seems to return a full list of apps.

Yes, same for me!

[...]

PR #372, opened 2020-05-07
updated mdfind command to match new call-format

xfirf:
Issue #308

### DIFF
--- a/src/main/executors/application-searcher.ts
+++ b/src/main/executors/application-searcher.ts
@@ -36,7 +36,7 @@ export function searchMacApplications(applicationSearchOptions: ApplicationSearc
                 return `^${applicationFolder.replace("\\", "\\\\").replace("/", "\\/")}`;
             }).join("|");
 
-            executeCommandWithOutput(`mdfind "kind:apps" | egrep "${folderRegex}"`)
+            executeCommandWithOutput(`mdfind kMDItemContentTypeTree=com.apple.application-bundle | egrep "${folderRegex}"`)
                 .then((data) => {
                     const filePaths = data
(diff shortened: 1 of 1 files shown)


oliverschwendener:
Can you please target the `dev` branch?

xfirf:
Yes. Sure.

oliverschwendener:
Accoriding to [this page](https://ss64.com/osx/mdfind.html) the current solution is correct. Is there somewhere an official documentation which states that your solution is the correct way to retrieve apps?

pthoelken:
When will this fix merged to public and released? The error is thill there on my macOS.

xfirf:
> According to [this page](https://ss64.com/osx/mdfind.html) the current solution is correct. Is there somewhere an official documentation which states that your solution is the correct way to retrieve apps?

Sorry for the delay. I´m not refering to the official documentation but to the comment of the issue #308

pthoelken:
@oliverschwendener maybe 
`mdfind kMDItemContentTypeTree=com.apple.application-bundle`
works on macOS 10.14.x either or did you test it already?

pthoelken:
To serve both versions of macOS in this command is maybe the option of an if-statement which check the macOS version and execute the command depends from the if-result which should pass to the macOS version.

oliverschwendener:
I would leave it as it is for older macOS versions and use the new format only for macOS 10.15 and newer.

pthoelken:
Okay, when we can get the update for macOS 10.15?

oliverschwendener:
Closing this PR because it is obsolete. See 7d6a7fc.

Issue #366, opened 2020-05-03
[Feature Request] Reminder support


theRealPadster:
I recently found Ueli and would love to switch from [Lacona](https://setapp.com/apps/lacona), as Ueli is open source and on GitHub, but I frequently use the Lacona reminders feature, and that's one thing Ueli is lacking. 

Basically, you type in "remind me to" or "rmt" for short, then a string for the reminder, and then "at (some time)". And then it pops up a notification at that time. 

e.g. "rmt drink water in 2 hours" or "rmt call mom tomorrow" or "rmt do the dishes at 18:00"

themegaphoenix:
Hello,
I have laid you the basic of the plugin with pull request #368 
it needs more testing but it seems to work in windows

theRealPadster:
Wow that was fast, thanks! I can try and test it some time this week on my MacBook

themegaphoenix:
@theRealPadster I would be grateful if you can

oliverschwendener:
I'm closing this issue as I have no plans to implement this. If anyone wants to implement this, feel free to reopen.

PR #374, opened 2020-05-07
Add Reminder plugin

themegaphoenix:
This will allow the user to set reminder
Only tested in windows 10! Needs testing in Mac and Windows 7/8
Icon taken from: https://www.iconfinder.com/icons/285670/calendar_icon
ParseReminder: allows to parse expression in to reminders
Node-Schedule: allows to schedule events
Notifications: Default Electron notifications plugin

The current messages it can parse are:
- remind me to send invites at 3pm tomorrow
- remind me to wish Jamie happy birthday on 8/28
- remind me in 10 minutes to change the laundry
- remind me that the oil needs changed on Oct 4
- remind me next Friday to water the plants

Issue: #366


oliverschwendener:
Thanks for your contribution! I'll have a look and leave some comments.

package.json
```diff
@@ -7,6 +7,9 @@
     "type": "git",
     "url": "https://github.com/oliverschwendener/ueli"
   },
+  "build": {
+    "appId": "ueli.com"
+  },
```
oliverschwendener, line 12: Why is this needed?

package.json
```diff
@@ -65,7 +68,10 @@
   },
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.10.1",
+    "@types/node-schedule": "^1.3.0",
     "bulma": "^0.8.0",
-    "bulma-extensions": "^6.2.7"
+    "bulma-extensions": "^6.2.7",
+    "node-schedule": "^1.3.2",
+    "parse-reminder": "^1.2.0"
```
oliverschwendener, line 75: Do we really need these 3rd party packages? Can't this be implemented without any extra dependencies?

src/common/config/reminder-options.ts
```diff
@@ -0,0 +1,9 @@
+export interface ReminderOptions {
+    isEnabled: boolean;
+    prefix:string;
```
oliverschwendener, line 3: Please pay attention to formatting. Here is a missing space.

src/common/icon/default-icons.ts
```diff
@@ -1,6 +1,6 @@
-import { Icon } from "./icon";
+import {Icon} from "./icon";
 
-import { IconType } from "./icon-type";
+import {IconType} from "./icon-type";
```
oliverschwendener, line 3: These changes are unnecessary.

src/common/translation/german-translation-set.ts
```diff
@@ -345,4 +345,8 @@ export const germanTranslationSet: TranslationSet = {
 
     controlPanel: "Systemsteuerung",
     controlPanelSettingsDescription: "Dieses Plugin erlaubt es dir, Elemente der Systemsteuerung schnell zu finden.",
+
+    reminder: "Erinnerung",
+    reminderPrefix: "Präfix",
+    reminderSettingsDescription: "Mit diesem Plugin können Sie Erinnerungen festlegen\n",
```
oliverschwendener, line 351: Why is `\n` neede here?

src/common/translation/russian-translation-set.ts
```diff
@@ -345,4 +345,8 @@ export const russianTranslationSet: TranslationSet = {
 
     controlPanel: "панель управления",
     controlPanelSettingsDescription: "Этот плагин позволяет быстро находить элементы панели управления.",
+
+    reminder: "напоминание",
+    reminderPrefix: "Префикс",
+    reminderSettingsDescription: "Этот плагин позволяет устанавливать напоминания\n",
```
oliverschwendener, line 351: Why is `\n` neede here?

src/main/main.ts
```diff
@@ -15,7 +15,7 @@ import { getProductionSearchEngine } from "./production/production-search-engine
 import { GlobalHotKey } from "../common/global-hot-key/global-hot-key";
 import { defaultGeneralOptions } from "../common/config/general-options";
 import { getErrorSearchResultItem } from "../common/error-search-result-item";
-import { FileHelpers } from "./../common/helpers/file-helpers";
+import { FileHelpers } from "../common/helpers/file-helpers";
```
oliverschwendener, line 18: This change is not related to the actual purpose of this PR.

src/main/main.ts
```diff
@@ -460,7 +460,7 @@ function mainWindowNeedsToBeTransparent(userConfigOptions: UserConfigOptions): b
         return true;
     }
 
-    return userConfigOptions.appearanceOptions.allowTransparentBackground === true;
+    return userConfigOptions.appearanceOptions.allowTransparentBackground;
```
oliverschwendener, line 463: This change is not related to the actual purpose of this PR.

src/main/main.ts
```diff
@@ -470,7 +470,7 @@ function getMainWindowBackgroundColor(userConfigOptions: UserConfigOptions): str
         return transparent;
     }
 
-    return userConfigOptions.appearanceOptions.allowTransparentBackground === true
+    return userConfigOptions.appearanceOptions.allowTransparentBackground
```
oliverschwendener, line 473: This change is not related to the actual purpose of this PR.

src/main/main.ts
```diff
@@ -757,6 +757,7 @@ app.on("ready", () => {
         quitApp();
     }
 });
+app.setAppUserModelId("ueli.com");
```
oliverschwendener, line 760: Can you explain why this is needed?

src/main/plugins/reminder-plugin/reminder-plugin.ts
```diff
@@ -0,0 +1,94 @@
+import {ExecutionPlugin} from "../../execution-plugin";
+import {PluginType} from "../../plugin-type";
+import {SearchResultItem} from "../../../common/search-result-item";
+import {TranslationSet} from "../../../common/translation/translation-set";
+import {UserConfigOptions} from "../../../common/config/user-config-options";
+import {ReminderOptions} from "../../../common/config/reminder-options";
+import {defaultReminderIcon} from "../../../common/icon/default-icons";
+import {join} from "path";
+// @ts-ignore
```
oliverschwendener, line 9: Can you explain why you added the `@ts-ignore` here?

src/main/main.ts
themegaphoenix: the reason that i added this is line is because electron notification doesn't show otherwise in windows:
https://www.electronjs.org/docs/tutorial/notifications#windows

package.json
themegaphoenix: ParseReminder: allows to parse expression in to reminders - it simplifies development as it allows to break down the reminder expression to 'what', 'when' and 'where' which can then be used for the notifications
Node-Schedule: allows to schedule events - you can use the date parsereminder date to schedule notifications

src/common/icon/default-icons.ts
themegaphoenix: This changes were made by WebStorm when formatting the file.
Should I change it back or keep it?

src/common/translation/german-translation-set.ts
themegaphoenix: Good catch!
It seems that when i copied and pasted from Google Translate it brought it with it.
I will fix it

src/common/translation/russian-translation-set.ts
themegaphoenix: Good catch!
It seems that when i copied and pasted from Google Translate it brought it with it.
I will fix it

src/main/main.ts
themegaphoenix: this the notification with the line:
image
this is without:
image
do you want to change it to something else or remove the line entirely?

src/main/plugins/reminder-plugin/reminder-plugin.ts
themegaphoenix: withouth that line i get:
`TS7016: Could not find a declaration file for module 'parse-reminder'. 'C:\Users\USERNAME\OneDrive\Desktop\ueli\node_modu
les\parse-reminder\index.js' implicitly has an 'any' type.
`
Do you know how to fix this? I tried googling but I seem unable to get a solution.
It is my first time using Typescript so i dont know the inside out this errors

Changing to `import {parseReminder} from "parse-reminder";` still gives me the same error

src/common/icon/default-icons.ts
oliverschwendener: Change it back.

src/common/icon/default-icons.ts
```diff
@@ -95,3 +95,8 @@ export const defaultBookmarkIcon: Icon = {
...
+
+export const defaultReminderIcon :Icon= {
```
oliverschwendener, line 99: Here is a missing whitespace. Please format the code like this:

``` typescript
const defaultReminderIcon: Icon = { ... }
```

src/common/icon/default-icons.ts
```diff
@@ -95,3 +95,8 @@ export const defaultBookmarkIcon: Icon = {
...
...c0 1.1 0.8954 2 2 2h6 2 6c1.105 0 2-0.9 2-2v-14c0-1.1-0.895-2-2-2h-6-2-6z" fill="#bdc3c7"/><path d="m5 3c-1.1046 0-2 0.8954-2 2v14c0 1.105 0.8954 2 2 2h6 2 6c1.105 0 2-0.895 2-2v-14c0-1.1046-0.895-2-2-2h-6-2-6z" fill="#ecf0f1" transform="translate(0 1028.4)"/><path d="m5 1031.4c-1.1046 0-2 0.9-2 2v3h18v-3c0-1.1-0.895-2-2-2h-6-2-6z" fill="#e74c3c"/><path d="m7 5.5a1.5 1.5 0 1 1 -3 0 1.5 1.5 0 1 1 3 0z" fill="#c0392b" transform="translate(.5 1028.4)"/><path d="m6 1c-0.5523 0-1 0.4477-1 1v3c0 0.5523 0.4477 1 1 1s1-0.4477 1-1v-3c0-0.5523-0.4477-1-1-1z" fill="#bdc3c7" transform="translate(0 1028.4)"/><path d="m7 5.5a1.5 1.5 0 1 1 -3 0 1.5 1.5 0 1 1 3 0z" fill="#c0392b" transform="translate(12.5 1028.4)"/><g fill="#bdc3c7"><path d="m18 1029.4c-0.552 0-1 0.4-1 1v3c0 0.5 0.448 1 1 1s1-0.5 1-1v-3c0-0.6-0.448-1-1-1z"/><path d="m5 1039.4v2h2v-2h-2zm3 0v2h2v-2h-2zm3 0v2h2v-2h-2zm3 0v2h2v-2h-2zm3 0v2h2v-2h-2z"/><path d="m5 1042.4v2h2v-2h-2zm3 0v2h2v-2h-2zm3 0v2h2v-2h-2zm3 0v2h2v-2h-2zm3 0v2h2v-2h-2z"/><path d="m5 1045.4v2h2v-2h-2zm3 0v2h2v-2h-2zm3 0v2h2v-2h-2zm3 0v2h2v-2h-2zm3 0v2h2v-2h-2z"/></g><rect fill="#c0392b" height="1" transform="translate(0 1028.4)" width="18" x="3" y="8"/></g></svg>`,
```
oliverschwendener, line 100: Here is a missing whitespace

oliverschwendener:
Can you please target the `dev` branch?

theRealPadster:
I have successfully built and ran it under macOS Catalina. 

It detected my reminder query (though it seems to only support "min/minutes/sec/seconds" and not "30 s/30 m". 

The first time a reminder fired off, it asked if I wanted to allow Electron to display reminders, which I granted. I then had to set another reminder, as the first one didn't show up.
image

When the second one showed up, it had the Electron icon, not UELI. There is also no close button, like some notifications have, but it's not a huge deal. 
image

This is the Lacona reminder notification, for reference. The snoozing isn't a big deal, but it is something that I noticed is present. 
image

package.json
stefnotch: It should be quite possibly to replace parse reminder with a built-in implementation, as its code is rather simple. Replacing chrono-js, which parse-reminder depends on seems to be a fair bit harder. https://github.com/bkeepers/parse-reminder/blob/master/index.js

Replacing node-schedule with a custom made solution should be rather simple, because it doesn't actually schedule a cronjob or anything. Instead, "scheduled jobs will only fire as long as your script is running".

themegaphoenix: @stefnotch i will look in replacing parse-reminder with a custom one based on it - but having a look at the chrono-js it is beyond my capabilities right now
regarding the node-schedule i think a timeout would work? it would have the same capabilities. How would we go about notification when the code is not running i.e. restart?

themegaphoenix:
> I have successfully built and ran it under macOS Catalina.
> 
> It detected my reminder query (though it seems to only support "min/minutes/sec/seconds" and not "30 s/30 m".
> 
> The first time a reminder fired off, it asked if I wanted to allow Electron to display reminders, which I granted. I then had to set another reminder, as the first one didn't show up.
> image
> 
> When the second one showed up, it had the Electron icon, not UELI. There is also no close button, like some notifications have, but it's not a huge deal.
> image
> 
> This is the Lacona reminder notification, for reference. The snoozing isn't a big deal, but it is something that I noticed is present.
> image

regarding the authorisation, i think when enabling a reminder plugin should also show up a new notification so that the user can authorise it? it would be a dirty trick 
for the icon i think because of the relative icon the path is causing problems.
To help me diagnose can you add this line after the `exectute` function starts line 70 (before the promise)
` console.log(join(__dirname,'../img/icons/win/icon-transparent.ico'));`
this would tell me what directory the file is running from in the mac and thus issue a fix

regarding the extra buttons i think it could be added:
https://www.electronjs.org/docs/api/structures/notification-action
not exactly sure about it though

package.json
stefnotch: I suppose it would be acceptable to assume that ueli is always running and only show reminders while ueli is running. Regarding an ueli restart, I think you pretty much have to write the reminders to a file or somehow persist them. But I'm not sure, so I'm pinging @oliverschwendener

themegaphoenix: @stefnotch yeah you are right about a file/sqlite database for the persist. and load them everytime the reminder load. 
Would a calendar integration be better for this?

theRealPadster:
> for the icon i think because of the relative icon the path is causing problems.
> To help me diagnose can you add this line after the `exectute` function starts line 70 (before the promise)
> ` console.log(join(__dirname,'../img/icons/win/icon-transparent.ico'));`
> this would tell me what directory the file is running from in the mac and thus issue a fix
> regarding the extra buttons i think it could be added:
> https://www.electronjs.org/docs/api/structures/notification-action
> not exactly sure about it though



It logs: `/Users/(username)/Sources/ueli/img/icons/win/icon-transparent.ico`

I took a look at the button action thing, but I don't know anything about developing with Electron or Typescript, and it didn't make much sense to me, so I'm not sure how much help I can be there.

themegaphoenix:
> > for the icon i think because of the relative icon the path is causing problems.
> > To help me diagnose can you add this line after the `exectute` function starts line 70 (before the promise)
> > ` console.log(join(__dirname,'../img/icons/win/icon-transparent.ico'));`
> > this would tell me what directory the file is running from in the mac and thus issue a fix
> > regarding the extra buttons i think it could be added:
> > https://www.electronjs.org/docs/api/structures/notification-action
> > not exactly sure about it though
> 
> It logs: `/Users/(username)/Sources/ueli/img/icons/win/icon-transparent.ico`
> 
> I took a look at the button action thing, but I don't know anything about developing with Electron or Typescript, and it didn't make much sense to me, so I'm not sure how much help I can be there.

thank you for this! Does the path exists and is the icon there? If the icon is there I have no clue why it doesnt show up

package.json
themegaphoenix: @stefnotch @oliverschwendener i have implemented my own parse reminder. However for the scheduler i am unsure how to do it. Settimout max value is around 24 days. Would this be sufficient?

theRealPadster:
The icon is there. I'm not sure if it's an issue with it being in .ico format or what, but it previews fine when I try and open the image.

themegaphoenix:
> The icon is there. I'm not sure if it's an issue with it being in .ico format or what, but it previews fine when I try and open the image.

i think this might be of interest:
https://stackoverflow.com/questions/37547559/cannot-change-electron-notification-icon-on-left

theRealPadster:
> > The icon is there. I'm not sure if it's an issue with it being in .ico format or what, but it previews fine when I try and open the image.
> 
> i think this might be of interest:
> https://stackoverflow.com/questions/37547559/cannot-change-electron-notification-icon-on-left
> refreshing the cache might help:
> https://stackoverflow.com/questions/11856766/osx-notification-center-icon


If I change the icon path to one of the mac/png ones, it shows up on the right side, but not the app icon. 
image

The app still identifies itself as Electron for some reason, which would be why the icon is Electron. 
image

I googled electron app-bundle-id and found [this](https://github.com/electron-userland/electron-builder/issues/368),but it didn't seem to do anything when I added this to the package.json and rebuilt it. 
```
"build": {
    "appId": "org.test.testing"
  },
```

I'll have to do some more poking around later and see what I can dig up.

themegaphoenix:
> > > The icon is there. I'm not sure if it's an issue with it being in .ico format or what, but it previews fine when I try and open the image.
> > 
> > 
> > i think this might be of interest:
> > https://stackoverflow.com/questions/37547559/cannot-change-electron-notification-icon-on-left
> > refreshing the cache might help:
> > https://stackoverflow.com/questions/11856766/osx-notification-center-icon
> 
> If I change the icon path to one of the mac/png ones, it shows up on the right side, but not the app icon.
> image
> 
> The app still identifies itself as Electron for some reason, which would be why the icon is Electron.
> image
> 
> I googled electron app-bundle-id and found [this](https://github.com/electron-userland/electron-builder/issues/368),but it didn't seem to do anything when I added this to the package.json and rebuilt it.
> 
> ```
> "build": {
>     "appId": "org.test.testing"
>   },
> ```
> 
> I'll have to do some more poking around later and see what I can dig up.

thank you very much.
I actually had those lines in package.json before in the earlier commits but i removed it because i added them for a previous dependency i used to make the notifications, but then i realised that electron already had it built in
if you can do this it would be helpful:
1. uninstall your current ueli installation
2. add the build line in package.json
3. run `yarn package` to create a installer
4. install the app

hopefully this will fix it

theRealPadster:
Ah, maybe it was because I was just running it from the command line and didn't actually build/install it. I did what you said and it seems to work now. It shows up as UELI, not Electron. The "icon" specified is still the one on the left, so we probably don't need it if it'll automatically use the UELI icon. 

image

themegaphoenix:
> Ah, maybe it was because I was just running it from the command line and didn't actually build/install it. I did what you said and it seems to work now. It shows up as UELI, not Electron. The "icon" specified is still the one on the left, so we probably don't need it if it'll automatically use the UELI icon.
> 
> image

good to hear that everything works!
i will change so to that it checks which OS it is running and add the icon if needed
maybe even extract it the notification in its own helper function
thank you very much

theRealPadster:
Were you able to make any progress @themegaphoenix ? I've been looking forward to switching over from Lacona ;P

themegaphoenix:
@theRealPadster unfortunately I have not had much time lately.
I extract the notfication in to its own helper function and checked for the OS to set the icon.
Please test it

theRealPadster:
> @theRealPadster unfortunately I have not had much time lately.
> I extract the notfication in to its own helper function and checked for the OS to set the icon.
> Please test it

Ah, that's fair. If there's anything I can do to help, let me know :)
I pulled from your reminder branch (66 commits) and ran a `yarn package`, and installed it. I have the plugin enabled but for some reason it's not showing the reminder action when I try and run it. Is it working for you in Windows? Maybe I did something wrong?

themegaphoenix:
> Any updates on this?

This branch is now really outdated, so I will have to cancel this PR. Unfortunately I won't be able to add the feature :(

PR #384, opened 2020-05-13
Add event emitter to update current results

AmrBinBashir:
This PR is gonna solve my other PR #283.
Basically I just created an event emitter and listen to its events in the main process, so whenever a plugin needs to update the results, we can just import the emitter and emit the event.

### DIFF
--- a/src/common/ipc-channels.ts
+++ b/src/common/ipc-channels.ts
@@ -36,4 +36,5 @@ export enum IpcChannels {
     checkForUpdate = " check-for-update",
     checkForUpdateResponse = "check-for-update-response",
     downloadUpdate = "download-update",
+    updateCurrentResults = "update-current-results"
 }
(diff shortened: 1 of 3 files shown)


oliverschwendener:
Why would a plugin need to update the results directly? In my opinion this leads to an unclean architecture.

AmrBinBashir:
In my other PR about adding suggestion to webSearchPlugin you didn't want the plugin to be dependent on axios call, with this PR , plugins that would need some kind of GET call can provide default search results and wait for the GET call to resolve then trigger the event to update the current results.

oliverschwendener:
The only way to get search results from a plugin is through the SearchEngine in `getSearchResults` which then calls `getSearchResults` on all execution plugins. This is the only way a plugin should communicate to the outside. The plugin should not care about updating search results in another place. I think you should rethink the way the websearch plugin should work with suggestions.

Issue #267, opened 2020-01-20
Color theme import and export bug.


AmrBinBashir:
Steps to produce:
1 - Export a color theme 
2 - Import a color theme and select the same file you just exported 
it says 'invalid color theme'.

PR #387, opened 2020-05-16
fix import color theme bug

AmrBinBashir:
#267


oliverschwendener:
This solution only works if the color theme name is the first property in the JSON object.

PR #388, opened 2020-05-16
ctrl+Enter-to-add-.com-and-go-to-url

AmrBinBashir:


src/main/main.ts
```diff
@@ -738,6 +738,12 @@ function registerAllIpcListeners() {
                 .catch((err) => logger.error(err));
         }
     });
+
+    ipcMain.on(IpcChannels.executeUrl, (event, userInput: string) => {
+        openUrlInBrowser(`${config.urlOptions.defaultProtocol}://${userInput}.com`)
+            .then(() => { /* do nothing */ })
+            .catch((err) => logger.error(err));
+    })
```
oliverschwendener, line 746: This should be handled by the UrlPlugin.

AmrBinBashir: But I can't access the UrlPlugin directly, maybe extract the logic of making a url search result item into a helper function ?

PR #390, opened 2020-05-16
Fixes #361 : Results Number mess up icon size

### LINKED ISSUE #361: Results Number makes Icons with long description smaller
image

I can try to fix it but I want to make the numbers on the left beside the icon, I find myself always looking at the icons so it makes sense that the numbers should be next to the Icon. If you agree I can make a PR to fix this issue and move the numbers to the left.

AmrBinBashir:
I fixed the bug #361 but I added some modifications:
1- I moved the numbers from right to left next to the icons.
2- Styled the numbers and added a vue transition to them. 

Annotation 2020-05-16 155217

These two changes would play nicely with my other PR #351 if it got merged.

### DIFF
--- a/styles/app.css
+++ b/styles/app.css
@@ -232,7 +232,7 @@ body {
     box-sizing: border-box;
     display: flex;
     flex-direction: column;
-    max-width: calc(100% - var(--search-results--item-icon-size));  
+    max-width: calc(100% - var(--search-results--item-icon-size) * 2);
 }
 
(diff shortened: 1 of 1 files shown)


oliverschwendener:
1) I think the numbers look better on the right side.
2) I think the current styling looks better and there is no need for a transition here.

AmrBinBashir:
Okay i will just fix the issue only.

Issue #297, opened 2020-02-17
Commandline plugin history


AmrBinBashir:
whenever a user executes a command with command line plugin, it should be added to a history list and next time the user activates command line plugin instead of showing only one result with current typed command, it should display the commands used before as well, kinda like how windows "run" work.

oliverschwendener:
Did you try <kbd>Shift</kbd><kbd>↑</kbd>?

AmrBinBashir:
@oliverschwendener yes i did but that just cycles through all of my search queries of the whole app not just the command line plugin and while at it, the history array for the whole app shouldn't hold duplicate values, cause some times i use the same search query over and over again and when i try to cycle through them it takes forever to get what i want.

PR #391, opened 2020-05-17
Commandline plugin history

AmrBinBashir:
#297

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -674,6 +674,15 @@ function registerAllIpcListeners() {
             });
     });
 
+    ipcMain.on(IpcChannels.commandlineClearHistory, () => {
+        searchEngine.clearCommandlineHistory()
+            .then(() => notifyRenderer(translationSet.successfullyClearedCommandlineHistory, NotificationType.Info))
+            .catch((err) => {
+                logger.error(err)
+                notifyRenderer(err, NotificationType.Error);
+            });
+    })
+
     ipcMain.on(IpcChannels.openDebugLogRequested, () => {
         logger.openLog()
             .then(() => { /* do nothing */ })
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -676,7 +676,10 @@ function registerAllIpcListeners() {
 
     ipcMain.on(IpcChannels.commandlineClearHistory, () => {
         searchEngine.clearCommandlineHistory()
-            .then(() => notifyRenderer(translationSet.successfullyClearedCommandlineHistory, NotificationType.Info))
+            .then(() => {
+                refreshIndexOfPlugin(PluginType.Commandline);
+                notifyRenderer(translationSet.successfullyClearedCommandlineHistory, NotificationType.Info);
+            })
             .catch((err) => {
                 logger.error(err)
                 notifyRenderer(err, NotificationType.Error);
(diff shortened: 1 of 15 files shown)


src/renderer/renderer.ts
```diff
@@ -175,6 +175,10 @@ const app = new Vue({
             ipcRenderer.send(IpcChannels.downloadUpdate);
         });
 
+        vueEventDispatcher.$on(VueEventChannels.commandlineClearHistory, () => {
+            ipcRenderer.send(IpcChannels.commandlineClearHistory)
+        })
```
oliverschwendener, line 180: The `renderer.ts` shouldn't care about a specific plugins. In my opinion, because this is not a generic solution it does not scale properly.

src/main/search-engine.ts
```diff
@@ -180,6 +181,10 @@ export class SearchEngine {
         return this.favoriteManager.clearExecutionLog();
     }
 
+    public clearCommandlineHistory(): Promise<void> {
+        return new ElectronStoreCommandlineHistoryRepository().clearAll();
+    }
```
oliverschwendener, line 186: The SearchEngine should also not care about specific plugins, because it does not know which specific plugins are in use. Also not a very scalable solution.

src/main/main.ts
```diff
@@ -674,6 +674,18 @@ function registerAllIpcListeners() {
...
+        searchEngine.clearCommandlineHistory()
+            .then(() => {
+                refreshIndexOfPlugin(PluginType.Commandline);
+                notifyRenderer(translationSet.successfullyClearedCommandlineHistory, NotificationType.Info);
+            })
+            .catch((err) => {
+                logger.error(err)
+                notifyRenderer(err, NotificationType.Error);
+            });
+    })
```
oliverschwendener, line 687: `main.ts` should not care about specific plugins. Also not scalable.

src/main/search-engine.ts
AmrBinBashir: This means I will have to remove the option to clear the plugin history.

oliverschwendener:
I appreciate your effort, but this solution is way too plugin specific. I would prefer something more generic, something that is not strictly tied to a specific plugin. And honestly I don't really see the need for this when you can just use <kbd>Shift</kbd><kbd>↑</kbd>.

AmrBinBashir:
@oliverschwendener  I regularly execute ping command, so every time I want to execute the command, either I have to scroll through my history which is a bad ux imo, or I will have to type the command again which still bad ux.

oliverschwendener:
Take the command line for example: let's say you execute `ping`. Then after that, you type in some other commands, let's say `ipconfig`, `nslookup` and `dir`. If you want to execute `ping` again you have to scroll through your entire history. I don't get where you see bad user experience there. What you are trying to solve is that you want to have a separate user input history for a specific plugin, in your case the command line plugin. This would be fine in my opinion, but with a more generic approach please. What if I wanted to have a sepearte user input history for other plugins? Then we would have to reimplement everything from scratch again for all types of plugins. This does not scale well.

AmrBinBashir:
Cycling through the whole app history to search for a command would be a bad ux cause if i execute `ping` then open multiple apps next then maybe execute `ipconfig` or whatever then if I want to execute `ping` again I will have to cycle through a lot of irrelevant history which might inclue apps and other plugnis. On the other hand if commandline has its own history I can just scroll thought its history maybe sort out the history as I type in too. 

I get that my approach wasn't the best, but the idea itself isn't that bad. Maybe create a history manager class that any plugin can use.

For now, keep the PR closed until a better solution comes around.

PR #392, opened 2020-05-17
filter out duplicates from input history manager

AmrBinBashir:
Having duplicates in userInput history manager is inconvenient, A lot of times I open an application like 5 times in a row and It makes no sense to cycle through the same 5 input history.

### DIFF
--- a/src/main/user-input-history-manager.test.ts
+++ b/src/main/user-input-history-manager.test.ts
@@ -15,6 +15,26 @@ describe(TestableUserInputHistoryManager.name, (): void => {
             expect(actual[0]).toBe(userInput);
             expect(manager.getIndex()).toBe(1);
         });
+
+        it("should only add an item if it's not the same as the previous", (): void => {
+
+            const manager = new TestableUserInputHistoryManager();
+
+            const userInput1 = "test1";
+            const userInput2 = "test2";
+
+            manager.addItem(userInput1);
+            manager.addItem(userInput1);
+            manager.addItem(userInput2);
+
+            const actual = manager.getHistory();
+
+            expect(actual.length).toBe(2);
+            expect(actual[0]).toBe(userInput1);
+            expect(actual[1]).toBe(userInput2);
+            expect(manager.getIndex()).toBe(2);
+
+        });
     });
 
     describe("getPrevious", (): void => {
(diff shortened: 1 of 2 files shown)


oliverschwendener:
Your solution filters out duplicates entirely, but what you actually want is to remove duplicates in a row.

AmrBinBashir:
I thought about just removing the duplicates in a row but if i am cycling through the history and encounter vscode for example, I don't need to see it again, if I wanted it , I would've used it the first time I encountered it while cycling through, what do you think ?

oliverschwendener:
I expect the input history to correctly represent the history. No magic. I would be fine with removing duplicates in a row. Or maybe this: add a new item to the history only if the previous item is not the same.

AmrBinBashir:
Okay, I changed it to only add history item if it's not the same as the last one.

oliverschwendener:
Can you add a test for this functionality?

AmrBinBashir:
Added them.

oliverschwendener:
Can you add a separate test case for this which describes this exact use case? Sorry for the hassle.

AmrBinBashir:
Done

src/main/user-input-history-manager.test.ts
```diff
@@ -15,6 +15,26 @@ describe(TestableUserInputHistoryManager.name, (): void => {
             expect(actual[0]).toBe(userInput);
             expect(manager.getIndex()).toBe(1);
         });
+
+        it("should only add an item if it's not the same as the previous", (): void => {
+
+            const manager = new TestableUserInputHistoryManager;
```
oliverschwendener, line 21: I think there are parantheses missing. It should be `const manager = new TestableUserInputHistoryManager();`

AmrBinBashir: Yeah, that was my bad.

src/main/user-input-history-manager.ts
```diff
@@ -8,7 +8,9 @@ export class UserInputHistoryManager {
     }
 
     public addItem(userInput: string): void {
-        this.history.push(userInput);
+        if (this.history[this.history.length - 1] !== userInput) {
```
oliverschwendener, line 11: What happens if `this.history` is an empty array? Doesn't it fail because it can't access the item with index `-1`?

AmrBinBashir: It would return `undefined` which is not equal to `userInput`, it works fine I tested it. I guess I can add one more check to see if the array is not empty.

PR #394, opened 2020-05-18
Add folder type to Shortcuts Plugin

### LINKED ISSUE #338: Add folder type to Shortcuts Plugin
I always want to access my user directory but I can't index it using "Simple Folder Search" cause that will add more user directories that I don't want. 

Right now, I can add the folder with the type of file path, but It would be convenient to add Folder type too and A default folder Icon like "Simple Folder Search" Icon.

AmrBinBashir:
#338


oliverschwendener:
In my opinion this is just an additional dropdown option for the same thing.

Issue #401, opened 2020-05-21
Results not displaying non-western characters


xfcnx:
Some of my directories (folders) have non-western characters in them, and those characters are not displayed properly in the search results; for instance:

![Screenshot_1](https://user-images.githubusercontent.com/65690120/82519341-64e13a00-9b5c-11ea-89d4-dfcaf002f5c1.png)

As a consequence, such folders won't open when launched via ueli. I'm not sure if similar issues have been reported and any solutions or workarounds are known. Many thanks for your time and help!

newgivi:
I have the same problem with the Russian language.
When I use it in the ueli search queries — everything is good:
![ueli_660tFZaNDU](https://user-images.githubusercontent.com/65721393/82568743-928baa80-9b7f-11ea-8523-9607fc244715.png)
But when I try to search with the Everything plugin, I get question marks:
![ueli_dgJWffYhba](https://user-images.githubusercontent.com/65721393/82568745-93244100-9b7f-11ea-8062-8d38a157475c.png)
It's weird because both Everything and ueli supports the Russian language.

tkohlmeier:
I found the cause for this and created a pull request (#402).

But there is one more issue, visible in your first screenshot @newgivi. Applications with "special" characters in their names don't have the right icons. I have been looking into this, too, and it seems to be an issue with node-powershell. I will create another pull request for that.

tkohlmeier:
Found the cause for the Everything problem, too. #403 should fix it.

oliverschwendener:
Fix should be available in the newest release.

PR #402, opened 2020-05-21
File Searcher should use UTF8 encoding

tkohlmeier:
This fixes issue #401.
File searcher must set the output encoding to UTF8.


oliverschwendener:
Thanks!

PR #405, opened 2020-05-23
Fix opening control panel items with non-ASCII characters

tkohlmeier:
There are encoding issues with node-powershell when the script contains non-ASCII characters, like umlauts (äöü). That's why control panel items like `Geräte und Drucker` could not be opened. Using `executeCommand` fixes the issue.


oliverschwendener:
Thanks!

PR #406, opened 2020-05-23
Fix Windows app icon generation for files with special characters

tkohlmeier:
The windows-app-icon-generator uses a Powershell script to generate the icons. The file names are passed to powershell with node-powershell's addCommand() method. These strings are encoded as UTF8. The problem is, that this is not the encoding Powershell expects. Hence Powershell can't find the files and won't create the icons. The result is that the search results are shown without a proper icon, as can be seen in issue #401 (first screenshot: https://github.com/oliverschwendener/ueli/issues/401#issuecomment-632117828).

I could not find a way to make Powershell accept UTF8 input but I found another solution: Passing the information as base64-encoded JSON string.

### DIFF
--- a/src/main/plugins/application-search-plugin/windows-app-icon-generator.ts
+++ b/src/main/plugins/application-search-plugin/windows-app-icon-generator.ts
@@ -43,16 +43,29 @@ function generateIcons(icons: Icon[], followShortcuts?: boolean): Promise<void> 
             noProfile: true,
         });
 
-        ps.addCommand(`Add-Type -AssemblyName System.Drawing`);
+        const iconsJsonBase64 = Buffer.from(JSON.stringify(icons)).toString('base64');
 
-        icons.forEach((icon) => {
-            ps.addCommand(`$fileExists = Test-Path -LiteralPath "${icon.inputFilePath}";`);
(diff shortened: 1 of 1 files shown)


oliverschwendener:
@tkohlmeier Can we replace node-powershell entirely here?

tkohlmeier:
That's what I tried first. [powershell](https://www.npmjs.com/package/powershell) actually worked. I looked into the code and found the difference that is responsible for this. node-powershell passes the script to powershell by [writing to stdin](https://github.com/rannn505/node-powershell/blob/b0a0f7f33f7c0b07351e875007e97a20d217b420/lib/Shell.js#L188). powershell instead passes the script [as a command line argument](https://github.com/IonicaBizau/powershell/blob/068cd3c0750cfee617e39240142d8872fa68503b/lib/index.js#L43). I don't know why this works. Maybe some automatic encoding conversion is done there...

The problem is, this does not work for very long scripts:

    Error: spawn ENAMETOOLONG
    at ChildProcess.spawn (internal/child_process.js:394:11)

I even experimented with this in a C# program. Writing utf8-encoded strings to powershell's process stdin results in exactly the same behavior. So a real solution would be either writing correctly encoded strings to stdin (which I think is not supported by node.js) or changing the encoding powershell expects as stdin (which I could not achieve).

Maybe someone finds a cleaner solution, but until then, the workaround with base64 works just fine.

PR #407, opened 2020-05-26
fix commandline plugin execution pending

### LINKED ISSUE #395: commandline plugin pends the execution of the app until shell is closed
The command line plugin pends the execution of the app until the shell is closed. What if I want to run 2 different commands in two seprate windows. 

commandline plugin bug

I think the issue happens because of this commit 	5a4fbd7c7db8ac43a2b7fd4b5211a14a24480a9d

AmrBinBashir:
Fixes #395


oliverschwendener:
Hmm, the problem here is that if there is an error occurring during the execution, we'll never catch the error.

AmrBinBashir:
Passing the logger to the constructor and logging the errors inside the plugin.

oliverschwendener:
@AmrBinBashir Did you test this? Does this acutally work for your case you mentioned here: #395?

AmrBinBashir:
Yeah it works.

PR #414, opened 2020-05-28
Added iTerm Support.

alchemistake:
Shell commands needs some sanitation to run correctly. But it seems like it is not the priority number one.

Currently deployed version doesn't have sanitation as well.


oliverschwendener:
Thanks for this! Can you please resolve the linting issues? You can run `yarn lint` locally to see the issues.

alchemistake:
👍

PR #419, opened 2020-06-01
fix results item overlay on transparent bg

### LINKED ISSUE #418: inactive results item overlay on transparent background bug
On transparent background, the overlay is really annoying.
image

AmrBinBashir:
Fixes #418 
The overlay was causing this issue with transparent background 
image
so I removed it and decreased the opacity a little bit which kinda similar to the intended look I guess but doesn't cause issues with transparent background.
image


oliverschwendener:
Looks good to me, thanks!

PR #430, opened 2020-06-10
add alpha slider to color picker and use hex8

### LINKED ISSUE #426: Color picker Alpha slider
I noticed that color picker modal only uses the `hex` value out of `chrome-picker` and the alpha slider is hidden in the css which makes using transparent colors really hard. I think the alpha slider should be  visible and the color-picker-component should use either `rgba` or `hex8` instead of just `hex`.

AmrBinBashir:
Closes #426

### DIFF
--- a/styles/settings.css
+++ b/styles/settings.css
@@ -204,12 +204,7 @@ html {
     justify-content: center;
 }
 
-.vc-chrome-hue-wrap {
-    margin-bottom: 0 !important;
-}
-
 /* Hide color picker's text input (hex, rgba, etc.) */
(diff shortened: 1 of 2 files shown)


styles/settings.css
```diff
@@ -205,11 +205,10 @@ html {
 }
 
 .vc-chrome-hue-wrap {
-    margin-bottom: 0 !important;
+    margin-bottom: 10px !important;
 }
```
oliverschwendener, line 209: This whole css rule can be removed now.

AmrBinBashir: okay

Issue #429, opened 2020-06-10
Calculator result with comma separating thousands


JoeTessier:
Possibility to have the result (and even the figures in the seach bar) with comma between thousands.
Ex/ 10*5,000=50,000

oliverschwendener:
See #144

PR #376, opened 2020-05-07
Add separators for large number

themegaphoenix:
In reference to issue #144

PR #431, opened 2020-06-10
 #429 thousand separator 

weiwang11:
Current implementation just removes all commas from input before doing calculations without affecting input. Could maybe check if number has correct thousands separator?

preview1
preview2


weiwang11:
Not sure why the input modifier isn't passing the build on appveyor. We can remove the edits on `calculator.ts` if needed since the thousands separator on the results should work.

oliverschwendener:
Is this a duplicate of #376 ?

weiwang11:
It looks like it, but my implementation is using the built in function to convert separate the numbers by thousands but only for the calculator.

I think using:
```
Intl.NumberFormat().format(Number(result));
```
is much simplier than the implementation mentioned.

PR #435, opened 2020-06-12
fix workflow&shortcut plugins pending execution

### LINKED ISSUE #433: Shortcut to windows terminal hangs UELI
After struggling with sometimes not being able to type anything in UELI, I've narrowed the problem to the new Windows Terminal.

## Steps to reproduce
1. Launch `windows terminal (Preview)` from UELI from a shortcut (type: "Commandline Tool", command: "wt")
2. Press the hotkey that invokes UELI
3. You cannot type anything
4. Close the `windows terminal (Preview)` window that you opened
5. Press the hotkey that invokes UELI
6. You can now type again

With a terminal window open, restarting UELI also fixes the problem, allowing you to use UELI freely.

Launching windows `windows terminal (Preview)` through another mechanism (without using UELI, or using ueli but choosing the UWP app instead of the shortcut) does not trigger the bug, and UELI remains usable.

I believe this is due to some weird interaction where launching a terminal window does not return control or does not inform UELI that the launch was successful. However, launching `wt` from `cmd` does not result in this bug, and `cmd` remains usable.

### LINKED ISSUE #462: Ueli stops working after starting a workflow *Bug*
Hello,

I generated a workflow to open up different programs and documents. 
It contains the starting up of a Batch file and different cmd-lines.
The workflow works fine.
The problem occurs after that.
Ueli can be accessed by hotkey, but I cannot enter anything into the box.

amrbashir:
Closes #433, Closes #462

### DIFF
--- a/src/main/plugins/shortcuts-search-plugin/shortcuts-search-plugin.ts
+++ b/src/main/plugins/shortcuts-search-plugin/shortcuts-search-plugin.ts
@@ -9,6 +9,7 @@ import { isValidIcon } from "./../../../common/icon/icon-helpers";
 import { OpenLocationPlugin } from "../../open-location-plugin";
 import { stringIsWhiteSpace } from "../../../common/helpers/string-helpers";
 import { getDefaultShortcutIcon } from "./shortcut-helpers";
+import { Logger } from "../../../common/logger/logger";
 
 interface ExecutionArgumentDecodeResult {
     shortcutType: ShortcutType;
@@ -22,6 +23,8 @@ export class ShortcutsSearchPlugin implements SearchPlugin, OpenLocationPlugin {
     private readonly filePathExecutor: (filePath: string, privileged: boolean) => Promise<void>;
     private readonly filePathLocationExecutor: (filePath: string) => Promise<void>;
     private readonly commandlineExtecutor: (command: string) => Promise<void>;
+    private readonly logger: Logger;
+
 
     constructor(
         config: ShortcutOptions,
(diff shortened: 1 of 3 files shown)


src/main/plugins/shortcuts-search-plugin/shortcuts-search-plugin.ts
```diff
@@ -22,19 +23,23 @@ export class ShortcutsSearchPlugin implements SearchPlugin, OpenLocationPlugin {
     private readonly filePathExecutor: (filePath: string, privileged: boolean) => Promise<void>;
     private readonly filePathLocationExecutor: (filePath: string) => Promise<void>;
     private readonly commandlineExtecutor: (command: string) => Promise<void>;
```
gabriel-peracio, line 25:
```suggestion
    private readonly commandlineExecutor: (command: string) => Promise<void>;
```

You have a typo here

AmrBinBashir: I saw this misspell and I didn't want to change it tbh since it was oliver's code.

AmrBinBashir: I guess no harm in fixing that tho. I will add a commit to fix it.

AmrBinBashir: This was oliver's code, I didn't want to change it but i guess there is no harm in fixing a misspell.

Issue #443, opened 2020-06-18
Chrome Bookmarks: Index additional roots


jondcoleman:
Currently, only the bookmarks bar is indexed. You can also have bookmarks in Other Bookmarks (maybe also additional roots?)

![image](https://user-images.githubusercontent.com/11843663/85031852-2899fd00-b14d-11ea-8542-c625394259f1.png)

oliverschwendener:
A fix for this is included in the latest release.

PR #444, opened 2020-06-18
index additional chrome bookmark routes

jondcoleman:
re: #443


jondcoleman:
Hmm, looks like I should have branched from dev instead? Is that correct @oliverschwendener?

oliverschwendener:
Yes.

jondcoleman:
Ok, thanks. I'll close this and open a new one.

PR #455, opened 2020-06-23
generate windows icons without shortcut arrow

amrbashir:
This PR extracts the icons from the shortcut's target (original file).
Some shortcuts like folders , urls or system components (ex: This PC), the icon can't be extracted. In that case I just extract the icon from the shortcut itself.

I have 150 apps ,shortcuts,bat files indexed. 90% of them got the icon without the shortcut arrow.


oliverschwendener:
I only get errors:

image

image

AmrBinBashir:
It is working fine for me, 
image

Out of 28 apps only 2 or 4 were created with a shortcut arrow.
image


I think you have some shortcuts of deleted apps that throws an Exception, I will push a new commit to fix it.

PR #459, opened 2020-06-26
update dictionary api

AmrBinBashir:
The dictionary has stopped working for a while now because of outdated api url.


oliverschwendener:
Thanks!

PR #460, opened 2020-06-27
Ensure file-path executing brings file to foreground

sashaweiss:
This PR changes the `file-path-executor` to always use Powershell/`osascript`, rather than using `shell.openItem` for non-privileged openings. This ensures that the executed file path is always brought to the foreground.

For reasons I don't fully understand, certain application types are not brought to the foreground when opened with `ueli`. For example, MacOS preference panes open when launched with `ueli`, but the current application retains focus and so they appear to open in the background. Another example is Firefox, which has the same behavior when cold-launched by `ueli`.

@nathanshelly and I tested using the Electron `shell.openItem` API in a test app to launch Firefox, and it did not exhibit this behavior. We then discovered that disabling "Hide window after execution" in the `ueli` settings fixed this behavior, so we suspect it may be some weird interplay between apps with certain accessibility types and how Electron handles window and app hiding. We've discovered in the past that Firefox sometimes uses non-standard attributes on its windows, which caused [trouble in another app](https://github.com/lwouis/alt-tab-macos/issues/360).

Rather than try and unwind why hiding the `ueli` window/app interferes with Firefox/preference panes, this change seems to ensure they are foregrounded always.

### DIFF
--- a/src/main/executors/file-path-executor.ts
+++ b/src/main/executors/file-path-executor.ts
@@ -1,33 +1,17 @@
 import { executeCommand } from "./command-executor";
-import { shell } from "electron";
 
 export function executeFilePathWindows(filePath: string, privileged: boolean): Promise<void> {
-    return privileged
-        ? executeFilePathWindowsAsPrivileged(filePath)
-        : openFile(filePath);
+    const command: string = privileged ?
+        `powershell -Command "& {Start-Process -Verb runas '${filePath}'}"` :
+        `powershell -Command "& {Start-Process '${filePath}'}"`;
+
+    return executeCommand(command);
 }
 
 export function executeFilePathMacOs(filePath: string, privileged: boolean): Promise<void> {
-    return privileged
-        ? executeFilePathMacOsAsPrivileged(filePath)
-        : openFile(filePath);
-}
+    const command: string = privileged ?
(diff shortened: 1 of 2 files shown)


sashaweiss:
One thing I'm noticing is that it does introduce a small slowdown in the execution speed, which I'm not sure there's a way around...

sashaweiss:
Revisited the strategy. In experimenting more, I found that disabling the "return focus to the previously-focussed app" workaround prevented this issue, although it obviously had the issue that if I did a search in `ueli` and then hit `ESC` or something focus was not returned to the previous app.

So, instead of always executing, which was too slow to be a viable solution, I've made it optional for the `hideMainWindow` method to return focus to the previously-focussed app. Then, in the case of "execute a file path" I pass false to disable the workaround, since focus will be passed to the executed file path anyway.

In my testing, focus is now passed correctly for Firefox/preference panes as well as for other apps. Do you think this is a safe strategy?

sashaweiss:
I'm now noticing that this seems to affect all system-built-in apps on Mac, like System Preferences, Messages, Calendar, Contacts, etc., which is an interesting piece of the puzzle.

sashaweiss:
Ok, now this fix only seems to work half the time. I'm going to close this PR and open an issue instead while I look into this more.

PR #461, opened 2020-06-28
add autocomplete to simple-folder-search-plugin

AmrBinBashir:


src/main/plugins/simple-folder-search-plugin/simple-folder-search-plugin.ts
```diff
@@ -72,6 +73,11 @@ export class SimpleFolderSearchPlugin implements SearchPlugin, OpenLocationPlugi
             resolve();
         });
     }
+    public autoComplete(searchResultItem: SearchResultItem): string {
+        return searchResultItem.executionArgument.endsWith(sep)
+            ? searchResultItem.executionArgument
+            : `${searchResultItem.executionArgument}${sep}`;
```
oliverschwendener, line 79: What if the `searchResultItem` is a file? I guess then the `sep` is not needed.

AmrBinBashir: Okay but That's is the same method from file-browser-plugin. Should I change that too or leave it ?

PR #475, opened 2020-07-13
group appearance options into sections for better accessibility

AmrBinBashir:
image


oliverschwendener:
Nice! I think if we make groups, they should have a title, no?

AmrBinBashir:
Yeah a title would be nice, I was also thinking about grouping color theme options too.

AmrBinBashir:
I added titles and grouped color theme options too. 

image

oliverschwendener:
Can you merge the current `dev` back to your branch? Thanks!

AmrBinBashir:
Done, I also update missing translations but they are all from google translate.

oliverschwendener:
There is a missing title in the last section of the appearance settings. I would put "General" or something similar in there.

image

AmrBinBashir:
Done

oliverschwendener:
Can you add more spacing under the title? I think it would look better.

image

AmrBinBashir:
I didn't want to add margin to the class since it is used in different places, so I used a bulma helper class.

oliverschwendener:
Nice!

PR #485, opened 2020-07-30
add hot reload for development

AmrBinBashir:
I find myself a lot of times making small changes and re-bundling the application every time which is inconvenient and takes a lot of time so I added hot reload.


oliverschwendener:
In my opinion this only adds more code to maintain that is only used for development and doesn't bring that much of a value. You can still use this when developing on your local machine without adding this to the version control, or even on your separate branch. But I won't include this in this repository.

PR #489, opened 2020-08-05
Only return focus on hiding main window when needed

sashaweiss:
`main.ts` contains a "workaround to return focus" when the main window is hidden. For some reason, this causes all MacOS built-in apps (e.g. Safari, Messages) as well as MacOS preference panes (OS settings pages) and weirdly enough Firefox to not be focused when launched via ueli.

This PR makes the `hideMainWindow` method take an argument that helps determine if we should invoke the workaround or not, and disables the workaround when on "file execution" to prevent those apps from not being focused on launch.

@nathanshelly

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -151,7 +151,7 @@ function calculateY(display: Electron.Display): number {
 
 function onBlur() {
     if (config.generalOptions.hideMainWindowOnBlur) {
-        hideMainWindow();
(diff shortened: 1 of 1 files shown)


sashaweiss:
I've been using a locally-built distribution of `ueli` based on a fork I have that brings in these changes. However, I went to test the latest release from the main repo, and found that this bug seems to no longer repro.

@oliverschwendener, I'm going to close this PR as a result.

PR #507, opened 2020-08-31
escape everything special characters

### LINKED ISSUE #503: [Feature Request] add more filtering options in File Search
Just found a bug in the everything command line integration under Windows:

Error in debug.log after trying: `es?ext:pdf size:>1MB`

```
Error: Command failed: cmd /c chcp 65001>nul && "C:\Program Files\Everything\es.exe" -max-results 24 ext:pdf size:>1MB
Zugriff verweigert

    at ChildProcess.exithandler (child_process.js:308:12)
    at ChildProcess.emit (events.js:210:5)
    at maybeClose (internal/child_process.js:1021:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:283:5)
```

Everything doc says:
>Use ^ to escape \, &, |, >, < and ^.

So the error is related to a lacking escape of the command line parameters. But I think this can be fixed very quickly.
Thank you in advance!

AmrBinBashir:
Fixes #503 

Before:
image

After:
image

### DIFF
--- a/src/main/plugins/everything-plugin/everything-searcher.ts
+++ b/src/main/plugins/everything-plugin/everything-searcher.ts
@@ -14,14 +14,14 @@ export function everythingSearcher(
     return new Promise((resolve, reject) => {
         const searchTerm = userInput.replace(everythingSearchOptions.prefix, "").trim();
         const utf8Encoding = "cmd /c chcp 65001>nul &&";
-        const command = `${utf8Encoding} "${everythingSearchOptions.pathToEs}" -max-results ${everythingSearchOptions.maxSearchResults} ${searchTerm}`;
+        const command = `${utf8Encoding} "${everythingSearchOptions.pathToEs}" -max-results ${everythingSearchOptions.maxSearchResults} ${getEverythingSearchTerm(searchTerm)}`;
         exec(command, (everythingError, stdout, stderr) => {
             if (everythingError) {
                 reject(everythingError);
             } else if (stderr) {
                 reject(stderr);
             } else {
-                const filePaths =  stdout.trim()
+                const filePaths = stdout.trim()
                     .split("\n")
                     .map((line) => normalize(line).trim())
                     .filter((f) => f !== ".");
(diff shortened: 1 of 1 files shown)


src/main/plugins/everything-plugin/everything-searcher.ts
```diff
@@ -14,14 +14,14 @@ export function everythingSearcher(
     return new Promise((resolve, reject) => {
         const searchTerm = userInput.replace(everythingSearchOptions.prefix, "").trim();
         const utf8Encoding = "cmd /c chcp 65001>nul &&";
-        const command = `${utf8Encoding} "${everythingSearchOptions.pathToEs}" -max-results ${everythingSearchOptions.maxSearchResults} ${searchTerm}`;
+        const command = `${utf8Encoding} "${everythingSearchOptions.pathToEs}" -max-results ${everythingSearchOptions.maxSearchResults} ${getEverythingSearchTerm(searchTerm)}`;
```
oliverschwendener, line 17: Would it also work if we just put the search term into double or single quotes?

For example like this:
```typescript
const command = `${utf8Encoding} "${everythingSearchOptions.pathToEs}" -max-results ${everythingSearchOptions.maxSearchResults} "${searchTerm}"`;
```

oliverschwendener:
Thanks for your effort! I left a comment above. And by the way don't worry about the breaking build, some tests on `dev` are currently failing.

src/main/plugins/everything-plugin/everything-searcher.ts
AmrBinBashir: This would disable the everything syntax and treat the query passed to es command line as one string.

After further reading of everything docs, if a users uses the es command line interface in the terminal, they have to escape these special characters manually. 

Now I am not sure if this PR is valid anymore, I mean without this PR, users can still escape these special characters like this.
image

While this works fine, it might be tedious a bit for users of the regular everything, they are not used to escape these characters.

AmrBinBashir: There is also escaping dashes and spaces using double quotes,

> Use double quotes to escape spaces and switches.

but if ueli escaped the spaces and dashes automatically there is no way for users to pass switches. 

So on one hand, ueli can escape these characters and the plugin would be just like the app, on the other hand, users have to manually escape these characters if they want, which would make the plugin just like the command line interface.

PR #520, opened 2020-09-22
Fixed fuse constructor

Roemer:
No idea why but vscode proposed that and it actually fixes the build.

### DIFF
--- a/src/main/plugins/filebrowser-plugin/filebrowser-plugin.ts
+++ b/src/main/plugins/filebrowser-plugin/filebrowser-plugin.ts
@@ -155,7 +155,7 @@ export class FileBrowserExecutionPlugin implements ExecutionPlugin, AutoCompleti
         let sortedResutls: SearchResultItem[] = [];
 
         if (searchTerm) {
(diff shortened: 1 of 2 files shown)


oliverschwendener:
This works when bundling but not when running the unit tests. To be honest, I have no idea why :/

Roemer:
Yes I saw the same. Very strange.

oliverschwendener:
I also tried to update the dependencies, but didn't work either.

AmrBinBashir:
I think I found the solution.
1. Rnable this flag in tsconfig 
	```js
	"esModuleInterop": true
	```
2. Remove every star import in the project and refactor it from this 
	```js
	import * as <imported_name>
	```
	to this
	```js
	import <imported_name>
 	```
	for example: `import * as Fuse from "fuse.js"` to `import Fuse from "fuse.js"`

oliverschwendener:
Thanks @AmrBinBashir!

oliverschwendener:
It's fixed in `dev`

PR #541, opened 2020-10-17
Customizable decimal separator

tkohlmeier:
This adds a setting for the decimal separator to general settings (as you suggested in #339). It is used by calculator and currency converter.

grafik

grafik

grafik


oliverschwendener:
Thanks for your contribution! I'll take a look at it asap and add comments if needed.

src/renderer/settings/general-settings-component.ts
```diff
@@ -525,6 +525,22 @@ export const generalSettingsComponent = Vue.extend({
...
+                                            :class="{ 'is-success': config.generalOptions.decimalSeparator == '.' }"
+                                            @click="config.generalOptions.decimalSeparator = '.'; updateConfig();">.</button>
+                                        <button class="button"
+                                            :class="{ 'is-success': config.generalOptions.decimalSeparator == ',' }"
+                                            @click="config.generalOptions.decimalSeparator = ','; updateConfig();">,</button>
+                                    </div>
+                                </div>
+                            </div>
+                        </div>
+
```
oliverschwendener, line 543: Is there a reason why you decided not to add a normal `select` with the options "," and "."?

tkohlmeier: I oriented on [Bootstrap radio buttons](https://getbootstrap.com/docs/4.0/components/buttons/#checkbox-and-radio-buttons) because of two advantages compared with a select: You see all options at once, without the need to open the dropdown. And you can change the setting with just one click. Sadly, Bulma does not support this natively, but the result is pretty much the same.

oliverschwendener: Fair enough.

Issue #539, opened 2020-10-16
An enhancement request to the annoying "Refreshing Indexes..."


walter6469:
![image](https://user-images.githubusercontent.com/70185618/96201298-de4d4f00-0f79-11eb-85fa-7ad880959509.png)
This is the most frustrating thing about UELI for me, the refresh timing is the default timing, but I think this could be made to be done in background and a refreshing text can pop up to the right side in the search bar like this 
![image](https://user-images.githubusercontent.com/70185618/96201422-27050800-0f7a-11eb-90e7-84ffde3552d3.png) But just replace the "Proceed?" with "Refreshing" and make that till the refreshing is done, we are still able to use the previously refreshed index and once the refreshing is done "Refreshing..." changes to "Refreshed!" (maybe in green color) and now we can search from the newly refreshed index. UELI is pretty much meant to simplify and increase workflows thus increasing productivity which it already does good but this is just an enhancement request so the most frustrating thing with UELI is resolved and I'd be really happy if something is done about this.
Thanks :)

oliverschwendener:
Thanks for your feedback! There is already a ticket covering this issue: #473

I'll close this because it's a duplicate.

PR #542, opened 2020-10-18
Refresh index in background

tkohlmeier:
This moves index refreshing to the background, as suggested in #539.
Instead of disabling user input, we simply show a hint:

grafik

I also changed index refreshing of the application search plugin a little so that new applications are not displayed before icons are also available.


oliverschwendener:
Doesn't this block the UI? The reason why I initially disabled the user input when running the index refresh was because in certian scenarios the index refresh was blocking the renderer process and the user couldn't enter anything. 🤔

tkohlmeier:
It is blocked, but only for a short time of the entire process. And, even while the GUI does not respond, the pressed keys are not discarded.

I created a gif which demonstrates it:

refreshIndex

oliverschwendener:
I don't think that this improves the user experience. In my opinion the best long term solution is to move the whole search engine logic and index to the renderer process. See this: https://github.com/oliverschwendener/ueli/issues/242

PR #546, opened 2020-10-19
Add tab completion to web search suggestions

tmueller:
The new search suggestion feature works really well. This PR adds tab completion to web search suggestions.

There is an ugly part in the `autoComplete` method. All `webSearchEngines` are searched and compared to a part of the `executionArgument` only to recreate the web search engines prefix. But since executionArguments are built from a specific web search engines configured url template, this works quite stable.

### DIFF
--- a/src/main/plugins/websearch-plugin/websearch-plugin.ts
+++ b/src/main/plugins/websearch-plugin/websearch-plugin.ts
@@ -7,8 +7,9 @@ import { WebSearchEngine } from "./web-search-engine";
 import { TranslationSet } from "../../../common/translation/translation-set";
 import { defaultWebSearchIcon } from "../../../common/icon/default-icons";
 import { isValidIcon } from "../../../common/icon/icon-helpers";
+import { AutoCompletionPlugin } from "../../auto-completion-plugin";
 
-export class WebSearchPlugin implements ExecutionPlugin {
+export class WebSearchPlugin implements ExecutionPlugin, AutoCompletionPlugin {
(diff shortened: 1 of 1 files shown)


src/main/plugins/websearch-plugin/websearch-plugin.ts
```diff
@@ -83,6 +84,20 @@ export class WebSearchPlugin implements ExecutionPlugin {
...
+            RegExp.$1 :
+            searchResultItem.executionArgument;
+
+        const foundWebSearchEngine = this.config.webSearchEngines.find((websearchEngine) => {
+            return websearchEngine.url.includes(searchUrl);
+        });
+
+        const prefix = foundWebSearchEngine ? foundWebSearchEngine.prefix : "";
+
+        return `${prefix}${searchResultItem.name} `;
```
oliverschwendener, line 98: Is the whitespace after `${searchResultitem.name}` intended?

oliverschwendener: To me it would make more sense to have the autocompletion result without it. 🤔

tmueller: It indeed is intended. The idea behind it is, that you can continue refining your search string without having to hit space again. If you wouldn't want to refine your search string, you wouldn't have used tab, but would have hit `enter` in the first place.

Additionally it is most useful for using duckduckgos bangs.

oliverschwendener: Ok I see, makes sense.

PR #553, opened 2020-10-23
Web search plugin improvements/fixes

tkohlmeier:
Some improvements and fixes for the web search plugin:

1. Descriptions of results contained URL-encoded texts:

    grafik

   This doesn't look nice and encoding is only necessary for the search URL. So I don't encode the texts anymore:

    grafik

2. If the search prefix was part of a suggestion, it was removed:

    grafik

    Now, the suggestion is correctly shown:

    grafik

3. There was an encoding-related problem with Google suggestions:

    grafik

    This could be fixed by setting a user agent...

    grafik

### DIFF
--- a/src/main/plugins/websearch-plugin/websearch-plugin.ts
+++ b/src/main/plugins/websearch-plugin/websearch-plugin.ts
@@ -55,8 +55,8 @@ export class WebSearchPlugin implements ExecutionPlugin {
 
                     for (const webSearchEngine of webSearchEngines) {
                         results.push({
-                            description: this.buildDescription(webSearchEngine, userInput),
-                            executionArgument: this.buildExecutionArgument(webSearchEngine, userInput),
+                            description: this.buildDescriptionFromUserInput(webSearchEngine, userInput),
+                            executionArgument: this.buildExecutionArgumentFromUserInput(webSearchEngine, userInput),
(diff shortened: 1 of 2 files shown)


src/main/plugins/websearch-plugin/websearch-plugin.ts
```diff
@@ -95,24 +95,32 @@ export class WebSearchPlugin implements ExecutionPlugin {
...
+        return this.translationSet.websearchDescription
+            .replace("{{websearch_engine}}", webSearchEngine.name)
+            .replace("{{search_term}}", searchTerm);
+    }
+
+    private getSearchTerm(webSearchEngine: WebSearchEngine, userInput: string, doNotEncode = false): string {
         let searchTerm = userInput.replace(webSearchEngine.prefix, "");
 
-        if (webSearchEngine.encodeSearchTerm) {
+        if (webSearchEngine.encodeSearchTerm && !doNotEncode) {
```
oliverschwendener, line 111: `!doNotEncode` is in my opinion a little overcomplicated. Not not conditions are quite hard to read, so I suggest to rename this variable to `skipEncoding` or something similar. What do you think?

tkohlmeier: You are right, skipEnoding sounds better.

PR #554, opened 2020-10-24
show edit button first in web search

ina6ra:
I want an edit button for web search at the first of the line.
It is inconvenient because you cannot press the edit button without scrolling.


oliverschwendener:
Thanks for your contribution! Because the edit/remove button could be shown first in other places too, I adjusted it directly on my branch.

oliverschwendener:
This will be included in the next release.

PR #615, opened 2021-01-02
Add 'to' keyword support for currency conversions

Axieum:
* Adds support for the use of the "to" keyword for currency conversions, e.g. `1 GBP to AUD`
* Fixes an issue where the input validator accepted whitespace, but the action was not set up to handle this whitespace

### DIFF
--- a/src/main/plugins/currency-converter-plugin/currency-converter-plugin.ts
+++ b/src/main/plugins/currency-converter-plugin/currency-converter-plugin.ts
@@ -25,12 +25,13 @@ export class CurrencyConverterPlugin implements ExecutionPlugin {
     }
 
     public isValidUserInput(userInput: string, fallback?: boolean | undefined): boolean {
+        const keywords = ["in", "to"];
         const words = userInput.trim().split(" ");
         if (words.length === 4) {
             try {
-                return !isNaN(this.getNumber(words[0]))
+                return keywords.includes(words[2].toLowerCase())
+                    && !isNaN(this.getNumber(words[0]))
                     && this.isCurrencyCode(words[1])
-                    && words[2] === "in"
                     && this.isCurrencyCode(words[3]);
             } catch (err) {
                 return false;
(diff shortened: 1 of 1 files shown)


src/main/plugins/currency-converter-plugin/currency-converter-plugin.ts
```diff
@@ -25,12 +25,13 @@ export class CurrencyConverterPlugin implements ExecutionPlugin {
     }
 
     public isValidUserInput(userInput: string, fallback?: boolean | undefined): boolean {
+        const keywords = ["in", "to"];
```
Axieum, line 28: I decided to declare the keywords up here, instead of inline, am happy to change if you think `["in", to"].includes(...)` would read better?

src/main/plugins/currency-converter-plugin/currency-converter-plugin.ts
```diff
@@ -25,12 +25,13 @@ export class CurrencyConverterPlugin implements ExecutionPlugin {
...
         const words = userInput.trim().split(" ");
         if (words.length === 4) {
             try {
-                return !isNaN(this.getNumber(words[0]))
+                return keywords.includes(words[2].toLowerCase())
+                    && !isNaN(this.getNumber(words[0]))
                     && this.isCurrencyCode(words[1])
-                    && words[2] === "in"
```
Axieum, line 33: Moved the keyword check to be the first condition, so that numbers are not parsed when not necessary

src/main/plugins/currency-converter-plugin/currency-converter-plugin.ts
```diff
@@ -99,7 +100,7 @@ export class CurrencyConverterPlugin implements ExecutionPlugin {
     }
 
     private buildCurrencyConversion(userInput: string): CurrencyConversion {
-        const words = userInput.split(" ");
+        const words = userInput.trim().split(" ");
```
Axieum, line 103: Fixes an issue where the validator was checking against a trimmed input, but this was not
image

oliverschwendener (approved):
Works as expected, thanks for the contribution!

Issue #638, opened 2021-01-31
'reboot' also should trigger a restart


abhijeetviswam:
After spending so long in front of unix command line interfaces, I almost always type 'reboot' instead of 'restart' for restarting the system. 

It would be nice if ueli allows 'reboot' command as well as an alternative to 'restart'

abhijeetviswam:
I don't disagree. Restart should be the option to be displayed.

All I am suggesting is that ueli accepts alternates as well. And reboot is a fairly regularly used term for restarting your PC. Same goes for power off, power down etc. even though the OS marks it as shut down. I am also against removing the default terminology of restart and replacing it with reboot, but providing an alternative wouldn't hurt.

In fact, this can be taken further. Command prompt would be populated in windows search if I search for "cmd", but doesn't in ueli. Or let's say you just can't remember the name of that music player app, you type in music and voila - you get spotify in the results.

This could be achieved with a mapper file with some terms being mapped to it's alternatives. Like the parent term restart being mapped to reboot, spotify being mapped to terms like music, player etc. And when the mapper terms are being searched, the results should populate parent terms as well. Better if the user is provided an option to create custom mappings as well.

All these might be out of scope for this issue, but just saying that features like these could make ueli even more joyful to use than what it is already now.

kr15uk:
+1 on reboot. Also please add "halt" for a shut down.

vishaluchil:
Now both "Restart" and "Reboot" trigger a suggests a system restart for windows. Did not include halt for shutdown as it is not a widely used term for shutdown, it could also be confused for sleep/hibernate.

PR #639, opened 2021-01-31
Add 'reboot' operating system command

abhijeetviswam:
Fix for #638


abhijeetviswam:
Closes #638

oliverschwendener:
See my comment in #638 .

Additionally I think this would cause confusion if let's say I have set my language to German and then search for "bot". Depending on the given fuzzyness setting from the Search Engine Settings I could get "Neustarten" as a search result when using the search term "bot" which is totally confusing and not transparent to the user.

Issue #604, opened 2020-12-15
Add support for Firefox bookmarks


Warden20:
I want to make the searches with firefox and see the bookmarks

oliverschwendener:
Currently not possible.

al-lad:
I'm also interested. Is it not possible because of Firefox itself or just not included in ueli yet ?

oliverschwendener:
Technically it's possible, but currently ueli does not support reading the firefox bookmarks.

FunctionDJ:
I'll take a look at implementing this.

tkohlmeier:
I already did. Just need to finalize it and create the pull request.

FunctionDJ:
Awesome!

Warden20:
Awesome!

PR #659, opened 2021-02-28
Add support for Firefox bookmarks

tkohlmeier:
As requested in #604, this adds support for Firefox bookmarks.

First, the default profile is determined (using `ini` package). If no default profile is found, the first profile is used.
Then the bookmarks are retrieved from `places.sqlite` (using `sqlite3` package).

Additionally, I enabled index refresh on enabling/disabling browser bookmarks plugin.


src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
```diff
@@ -0,0 +1,61 @@
...
+        this.userDataFolderPath = userDataFolderPath;
+    }
+
+    public getBrowserBookmarks(): Promise<BrowserBookmark[]> {
+        return this.getDatabaseFilePath()
+            .then(dbFilePath => this.getBookmarks(dbFilePath));
+    }
+
+    private getDatabaseFilePath(): Promise<string> {
+        const profilesIniFilePath = this.userDataFolderPath + path.sep + 'profiles.ini';
```
oliverschwendener, line 31: Could we use backticks here?

```typescript
const profilesIniFilePath = `${this.userDataFolderPath}${path.sep}profiles.ini`; 
```

src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
```diff
@@ -0,0 +1,61 @@
...
+            .then(fileContent => {
+                const profiles: { [name: string]: { Default: string, Path: string, IsRelative: string } } = ini.parse(fileContent);
+
+                let profile = Object.values(profiles).find(p => p.Default === '1');
+                if (!profile) {
+                    profile = Object.values(profiles).find(p => p.Path);
+                }
+                if (profile) {
+                    const absoluteProfilePath = this.getAbsoluteProfilePath(profile);
+                    resolve(absoluteProfilePath + path.sep + 'places.sqlite');
```
oliverschwendener, line 42: Also here

src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
```diff
@@ -0,0 +1,61 @@
...
+        type: IconType.SVG
+    };
+
+    private readonly userDataFolderPath: string;
+
+    constructor(userDataFolderPath: string) {
+        this.userDataFolderPath = userDataFolderPath;
+    }
```
oliverschwendener, line 23: This could be simplified to this:

```typescript
constructor(private readonly userDataFolderPath: string) {}
```

src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    public getBrowserBookmarks(): Promise<BrowserBookmark[]> {
+        return this.getDatabaseFilePath()
+            .then(dbFilePath => this.getBookmarks(dbFilePath));
```
oliverschwendener, line 27: Is there a reason why we don't have error handling here?

I personally use firefox and when I check out this branch I get this error:

```
Error: SQLITE_ERROR: no such table: moz_bookmarks
```

oliverschwendener: Looks like it got the wrong profile 🤔

src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
tkohlmeier: done

src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
tkohlmeier: done

src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
tkohlmeier: Of course. I'm glad you [changed your mind](https://github.com/oliverschwendener/ueli/pull/250#discussion_r358374508). 😉

src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
tkohlmeier: Hmm... How does your `profiles.ini` look like? It's located in `%appdata%\Mozilla\Firefox`.

Mine looks like this and that's working fine:

    [Profile1]
    Name=[...]
    IsRelative=1
    Path=[...]
    Default=1

    [Profile0]
    Name=default
    IsRelative=1
    Path=[...]

    [General]
    StartWithLastProfile=1
    Version=2

    [InstallE7CF176E110C211B]
    Default=[...]
    Locked=1

src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
oliverschwendener: Haha, yeah sorry about that one 😅

src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
oliverschwendener:
```
[Profile1]
Name=default
IsRelative=1
Path=[...]
Default=1

[Profile0]
Name=default-release
IsRelative=1
Path=[...]

[General]
StartWithLastProfile=1
Version=2

[Install2656FF1E876E9973]
Default=[...]
Locked=1
```

When I add/remove/modify bookmarks in the browser I can see that the `places.sqlite` file in the [Profile0] folder is being touched.

tkohlmeier: Oh, that's probably because of `StartWithLastProfile=1`. So you normally don't use your default profile. I will try to find out where the last used profile is stored for that case.

tkohlmeier: According to [this page](http://kb.mozillazine.org/Profiles.ini_file), `StartWithLastProfile=1` instructs Firefox to start using the profile with `Default=1` without showing the profile manager. This information seems to be outdated. When I changed the default profile using the profile manager, the Default=1 flag did not change, but the path under `[Install2656FF1E876E9973].Default` changed as well as the path in `installs.ini`. I'm going to change the default profile detection to use this setting.

tkohlmeier: Now using the profile whose path matches `[Install2656FF1E876E9973].Default`. As a fallback the profile with `Default=1` is used and else just the first profile.

src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    private getAbsoluteProfilePath(profilesIni: { [key: string]: any }): string | undefined {
+        const profiles: { Default: string, Path: string, IsRelative: string }[] = Object.keys(profilesIni)
+            .filter(iniSection => iniSection.startsWith('Profile'))
+            .map(iniSection => profilesIni[iniSection]);
+
+        let profile: { Default: string, Path: string, IsRelative: string } | undefined;
+        if (profilesIni.InstallE7CF176E110C211B?.Default) {
+            profile = profiles.find(p => p.Path === profilesIni.InstallE7CF176E110C211B.Default);
```
oliverschwendener, line 48: This seems very hardcoded here.

I suggest something like this:

```typescript
const installKey = Object.keys(profilesIni).find((iniSection) => iniSection.startsWith("Install"));

if (installKey && profilesIni[installKey]) {
    profile = profiles.find((p) => p.Path === profilesIni[installKey].Default);
}
```

src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
```diff
@@ -0,0 +1,69 @@
...
+            profile = profiles.find(p => p.Default === '1');
+        }
+        if (!profile) {
+            profile = profiles.find(p => p.Path);
+        }
+        if (!profile?.Path) {
+            return undefined;
+        }
+        return profile.IsRelative
+            ? this.userDataFolderPath + path.sep + profile.Path.replaceAll('/', path.sep)
```
oliverschwendener, line 60: Can we use template strings here too?

src/main/plugins/browser-bookmarks-plugin/firefox-bookmark-repository.ts
```diff
@@ -0,0 +1,69 @@
...
+        return new Promise((resolve, reject) => FileHelpers.readFile(profilesIniFilePath)
+            .then(fileContent => {
+                const profilesIni = ini.parse(fileContent);
+                const profilePath = this.getAbsoluteProfilePath(profilesIni);
+                if (profilePath) {
+                    resolve(`${profilePath}${path.sep}places.sqlite`);
+                } else {
+                    reject('Profile not found');
+                }
+            })
```
oliverschwendener, line 37: Do we need error handling here?

tkohlmeier: I'm not exactly sure what you mean. I would discourage from resolving this promise in case something goes wrong. If someone enables the browse bookmarks plugin and chooses Firefox, an error should be logged if something did not work.

oliverschwendener: Ah yes you're right. I was thinking about adding a `catch` statement here, but it would be unnecessary.

oliverschwendener:
Nice work, thanks alot!

Issue #644, opened 2021-02-14
[Feature Request] Allow Powershell in Workflow plugin


FunctionDJ:
Right now the options for execution steps are Command line tool, File path and URL. It would be great to be able to run Powershell as well in order to use modern Powershell commands instead of the limited cmd.

oliverschwendener:
In ueli v9 you are able to open Powershell with a given command:

![image](https://github.com/oliverschwendener/ueli/assets/15727229/5098f15e-6e0b-4b1b-96a4-fba391d6cf86)

PR #670, opened 2021-03-23
Execute workflow commands in default terminal

underthecocotree:
Hi,

While working with the Workflows I noticed that the commands where not firing in the terminal.  Maybe it is the kind of commands I wanted to run - but not much seemed to be happening for me.

The workflow command is now mimicking the commandline-plugin.  This change has been tested with iTerm on a Mac.  I noticed that if there are a few commands back to back they will execute on the same window. 

Related?
https://github.com/oliverschwendener/ueli/issues/644

### DIFF
--- a/src/main/plugins/workflow-plugin/workflow-plugin.ts
+++ b/src/main/plugins/workflow-plugin/workflow-plugin.ts
@@ -8,21 +8,26 @@ import { WorkflowExecutionStep } from "./workflow-execution-argument";
 import { WorkflowExecutionArgumentType } from "./workflow-execution-argument-type";
 import { Workflow } from "./workflow";
 import { defaultWorkflowIcon } from "../../../common/icon/default-icons";
+import { WindowsShell, MacOsShell } from "../commandline-plugin/shells";
+import { CommandlineOptions } from "../../../common/config/commandline-options";
 
 export class WorkflowPlugin implements SearchPlugin {
     public pluginType = PluginType.Workflow;
     private config: WorkflowOptions;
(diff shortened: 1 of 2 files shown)


oliverschwendener:
Cool, thanks for your contribution! I'll have a look at this as soon as I have time 👍

src/main/plugins/workflow-plugin/workflow-plugin.ts
```diff
@@ -8,21 +8,26 @@ import { WorkflowExecutionStep } from "./workflow-execution-argument";
 import { WorkflowExecutionArgumentType } from "./workflow-execution-argument-type";
 import { Workflow } from "./workflow";
 import { defaultWorkflowIcon } from "../../../common/icon/default-icons";
+import { WindowsShell, MacOsShell } from "../commandline-plugin/shells";
+import { CommandlineOptions } from "../../../common/config/commandline-options";
 
 export class WorkflowPlugin implements SearchPlugin {
     public pluginType = PluginType.Workflow;
     private config: WorkflowOptions;
+    private commandConfig: CommandlineOptions;
```
oliverschwendener, line 17: Can we rename this to `commandlineOptions`? I think it would be a more accurate name.

oliverschwendener (changes requested):
Thanks for the efforts, it works as expected. The only thing that I would like to be adressed is mentioned in the comment above.

underthecocotree:
@oliverschwendener just updated it.  I did battle with that variable name but now it makes a lot of sense.

oliverschwendener (approved):

PR #735, opened 2021-06-06
Add a Korean translation.

ddarkr:
Hello :) I love this application! but there is no Korean language, so I translated it.

 - It translated based on English, and there some liberal translation.
 - Checked application work after bundle
 - I set up vscode + tslint because I didn't use that, but I don't know if it worked properly. I'd appreciate it if you could check it out.

Translation suggestions and reviews are welcome.

Thank you for your good work.


oliverschwendener (approved):
Thanks for your contribution! Looks good to me

PR #736, opened 2021-06-07
Add Japanese translation

ykhrito:
I've just added Japanese translation.
Also I found small bug in `getItemName` for setting menu item, so fixed it.

Thank you for the great job!


oliverschwendener (approved):
Thanks alot!

PR #763, opened 2021-07-05
Reboot as alternative to restart #638

vishaluchil:
Now "reboot" and well as "restart" suggest Restart as the system command. Did not do the "halt" for shut down as it is not a term that is widely used and it could also mean sleep/hibernate for others


vishaluchil:
Sorry about the mistake. The two commits refer to different issues

PR #764, opened 2021-07-06
"Reboot" triggers system restart #638

vishaluchil:
Now both "Restart" and "Reboot" trigger a suggests a system restart for windows. Did not include halt for shutdown as it is not a widely used term for shutdown, it could also be confused for sleep/hibernate


vishaluchil:
Closing. Will follow up with new PR based on new branch.

PR #765, opened 2021-07-06
Added a more relevant URL icon

### LINKED ISSUE #558: [Feature request] Change Icon of "Open with web browser"
Screenshot_3

See in the screenshot, the icon for the "Open with web browser" command looks like opening some folder. Maybe change the icon to the image on the left in the screenshot?

:)

vishaluchil:
Now it looks like this:

image

Resolves #558


oliverschwendener:
Why do you think this is a more accurate icon?

vishaluchil:
I know that the previous one is supposed to be an arrow pointing to a browser window (Which I also found out while looking for icons). But I think the previous icon looks more like a folder icon on a small scale. Also, my initial intention was to add the globe icon as mentioned in the issue #558 but I couldn't find one that was free and required no attribution

vishaluchil:
Also, why I think this icon is more relevant is because it is most associated icon when talking about links in general. So thought this might be a more relevant icon for url and web search plugins

vishaluchil:
If you want, I can change it to this icon:
image

Is this okay?

oliverschwendener:
Alright, the globe is fine for me 👍

vishaluchil:
Logo changed to globe icon

PR #766, opened 2021-07-06
"Reboot" also triggers system restart, resolves #638

### LINKED ISSUE #638: 'reboot' also should trigger a restart

vishaluchil:
Now both "Restart" and "Reboot" trigger a suggests a system restart for windows. Did not include halt for shutdown as it is not a widely used term for shutdown, it could also be confused for sleep/hibernate, resolves #638

### DIFF
--- a/src/common/translation/translation-set.ts
+++ b/src/common/translation/translation-set.ts
@@ -260,6 +260,8 @@ export interface TranslationSet {
     windowsShutdownDescription: string;
     windowsRestart: string;
     windowsRestartDescription: string;
+    windowsReboot: string;
+    windowsRebootDescription: string;
     windowsSignout: string;
     windowsSignoutDescription: string;
     windowsLock: string;
--- a/src/common/translation/translation-set.ts
+++ b/src/common/translation/translation-set.ts
@@ -261,7 +261,6 @@ export interface TranslationSet {
     windowsRestart: string;
     windowsRestartDescription: string;
     windowsReboot: string;
-    windowsRebootDescription: string;
     windowsSignout: string;
     windowsSignoutDescription: string;
     windowsLock: string;
--- a/src/main/plugins/operating-system-commands-plugin/windows-operating-system-command-repository.ts
+++ b/src/main/plugins/operating-system-commands-plugin/windows-operating-system-command-repository.ts
@@ -44,7 +44,7 @@ export class WindowsOperatingSystemCommandRepository implements OperatingSystemC
                     type: IconType.SVG,
                 },
                 name: this.translationSet.windowsRestart,
-                searchable: [this.translationSet.windowsRestart, "reboot"],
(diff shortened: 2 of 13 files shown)


src/main/plugins/operating-system-commands-plugin/windows-operating-system-command-repository.ts
```diff
@@ -44,7 +44,7 @@ export class WindowsOperatingSystemCommandRepository implements OperatingSystemC
                     type: IconType.SVG,
                 },
                 name: this.translationSet.windowsRestart,
-                searchable: [this.translationSet.windowsRestart],
+                searchable: [this.translationSet.windowsRestart, "reboot"],
```
oliverschwendener, line 47: What if I have the language set to something else than English? ;)

vishaluchil: For most languages, reboot and restart have the same word. For some that have different words, would it be a good idea to create a file with an array and import and use it here. 

OR 

Since this is specific issue related to only restart/reboot and not any other system commands, would you like me to create that array in this file itself?

vishaluchil: I think I get what you mean. Eg. If I have it set to czech, then reboot should not suggest restart since there isn't a word like "reboot" in czech.

I think the only solution to that is to add a windowsReboot to the `translation-set.ts` and change only the searchable to:
```suggestion
                searchable: [this.translationSet.windowsRestart, this.translationSet.windowsReboot],
```

vishaluchil:
This works for all languages now

src/common/translation/translation-set.ts
```diff
@@ -260,6 +260,8 @@ export interface TranslationSet {
     windowsShutdownDescription: string;
     windowsRestart: string;
     windowsRestartDescription: string;
+    windowsReboot: string;
+    windowsRebootDescription: string;
```
oliverschwendener, line 264: I think the `windowsRebootDescription` is not used anywhere so I suggest that we remove it.

oliverschwendener (changes requested):
Apart from my comment the code looks good, thanks alot!

src/common/translation/translation-set.ts
vishaluchil: Yeah, you're right. I guess I just did it that way for it to look more consistent, but yes, it isn't useful at all. Will make the changes right away

vishaluchil:
Removed all occurrences of windowsRebootDesription. Everything should be good now

oliverschwendener:
Cool, thanks alot for your efforts! 💪

PR #767, opened 2021-07-07
Also works without "in" or "to" keywords

vishaluchil:
The first word after the number is treated as the base currency and the second is treated as the target currency.

image


vishaluchil:
Any update on this PR?

oliverschwendener:
Looks good, thanks alot!

PR #780, opened 2021-08-10
Add Shades of Purple Theme

Khoding:
Add Shades Of Purple (based more or less on https://github.com/ahmadawais/shades-of-purple-vscode, tested a few things to make it look as good as possible)


oliverschwendener:
Thanks alot!

Issue #665, opened 2021-03-07
Search engine search can miss last characters entered


FunctionDJ:
This issue exists in the latest stable version, details at the bottom.

**When typing at medium or higher speed, the search results and resulting URLs can miss the last character or even two from the search query.**

It's kind of random, but can be reproduced with a handful of retries.
I have a feeling it's related to the results becoming fewer when you type, which might block the thread from processing the key event. If that was true, it could be fixed by always processing the search input again after updating the results.

Screenshot: (unless you continue to type, it stays this way. resulting URLs also miss the characters)
![image](https://user-images.githubusercontent.com/34871211/110248589-378e3d00-7f72-11eb-826f-8298a3d6bf66.png)

Versions:
![image](https://user-images.githubusercontent.com/34871211/110248778-167a1c00-7f73-11eb-814f-3a69b1e54ec9.png)

edoreld:
I'm encountering the same bug.

How to reproduce:
1. Invoke ueli
2. Quickly type "d? duck" followed by the Return key
3. DuckDuckGo search opens with "du" as the search query

Observations:
 - There's input lag when typing on ueli, meaning that characters appear on ueli slower than I type them on my keyboard. 
 - When I type "d? duck <return>", only "du" appears on ueli. 

Possible solutions:
 - Remove input lag from ueli
 - Prevent the browser from being launched before the text has been processed

I could try to take a look if someone could point in the right way.

edoreld:
When entering an URL in ueli, the same problem doesn't appear.

If I enter "google.com" + return, it always takes me google.com (and not, for example, google.c)

Why is there input lag in one place but not in another? Maybe that's a hint of how to solve the issue.

oliverschwendener:
The input lag is because when you enter a search term in addition to the webesarch engine prefix ueli will call the search engines API to get the suggestions. When you're faster with typing than the HTTP request takes to respond, then this issue occurs. When you just enter the URL there is no need to do HTTP requests and wait for the response.

edoreld:
As a temporary solution, one can edit the web search that causes the problem and empty the "Suggestion URL" field. This will prevent suggestions from being loaded.

FunctionDJ:
> As a temporary solution, one can edit the web search that causes the problem and empty the "Suggestion URL" field. This will prevent suggestions from being loaded.

This is a good idea. I love that ueli shows suggestions, but i'd rather give up that, than live with the bug :D

firinael:
[...]

PR #788, opened 2021-09-04
Display search results only if they were retrieved for last user input

### LINKED ISSUE #778: [Bugs] Browser search can't keep up with typing
I typed an 8 letter word, ueli just registered 1 letter.
7osnGoOIN4

Browser search plugin doesnt register the whole of what I type. In order for it to properly work I have to slowly type, even then sometimes it will have a bug where some letters will not register, for example:

NVIDIA_Share_ygZT42dc1a
In this case ueli didnt register the last 6 letters of the search. Sometimes this will happen to be the letters in the middle, or beginning; it is random but mostly happens to the last letters.

These bugs are really hampering with my productivity, please fix them.

Thanks.

tkohlmeier:
Discard search results if they were not retrieved for the latest user input.

This should fix #778 and #665.

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -86,6 +86,7 @@ let trayIcon: Tray;
 let mainWindow: BrowserWindow;
 let settingsWindow: BrowserWindow;
 let lastWindowPosition = config.generalOptions.lastWindowPosition;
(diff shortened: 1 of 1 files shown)


oliverschwendener:
Looks good to me, thanks alot! 💪

PR #808, opened 2021-10-01
Got rid of for-loop using .findIndex

NotWearingPants:


src/main/favorites/electron-store-favorite-repository.ts
```diff
@@ -42,13 +42,8 @@ export class ElectronStoreFavoriteRepository implements FavoriteRepository {
...
     public update(favorite: Favorite): void {
-        // tslint:disable-next-line: prefer-for-of because we need
-        for (let i = 0; i < this.favorites.length; i++) {
-            if (this.favorites[i].item.executionArgument === favorite.item.executionArgument) {
-                this.favorites[i] = favorite;
-                break;
-            }
-        }
+        const index = this.favorites.findIndex((f) => f.item.executionArgument === favorite.item.executionArgument);
+        this.favorites[index] = favorite;
```
oliverschwendener, line 46: Thanks for you contribution! In your implementation `index` is possibly `-1` so we have to do a check for that before line 46.

What do you think about this approach? In my opinion it would simplify it even more.

```typescript
public update(favoriteToUpdate: Favorite): void {
    this.favorites = this.favorites.map((favorite) =>
        favorite.item.executionArgument === favoriteToUpdate.item.executionArgument ? favoriteToUpdate : favorite,
    );

    this.store.set(this.favoritesStoreKey, this.favorites);
}
```

oliverschwendener:
And please target the `dev` branch.

src/main/favorites/electron-store-favorite-repository.ts
NotWearingPants: Personally I prefer the index version, since it's more obvious that it changes only a single item which is the one specified, and also we can throw if the favorite wasn't in the list (if `-1`) instead of silently ignoring what was probably an error.

But the approaches are pretty much the same and the map one keeps the array immutable which can be a plus even though it copies everything.

I'll change to the map since you prefer it and I like it either way 👍

oliverschwendener (approved):

PR #810, opened 2021-10-01
Used Array.map's index instead of counting separately

NotWearingPants:


src/renderer/search-results-component.ts
```diff
@@ -43,13 +43,8 @@ export const searchResultsComponent = Vue.extend({
...
         },
         update(searchResults: SearchResultItem[]) {
-            let counter = 0;
-
             const viewModel = searchResults.map(
-                (searchResult): SearchResultItemViewModel => {
-                    counter++;
-                    return createSearchResultItemViewModel(searchResult, counter);
-                },
+                (searchResult, counter): SearchResultItemViewModel => createSearchResultItemViewModel(searchResult, counter),
```
oliverschwendener, line 47: In your implementation counter is `0` for the first item and before the count started at `1`.

oliverschwendener (changes requested):
See my comment.

oliverschwendener:
And please target the `dev` branch.

src/renderer/search-results-component.ts
oliverschwendener: Here is what I would suggest:

```typescript
const viewModel = searchResults.map(
    (searchResult, index): SearchResultItemViewModel =>
        createSearchResultItemViewModel(searchResult, index + 1),
);
```

NotWearingPants: Whoops, fixed :)

oliverschwendener (approved):

PR #812, opened 2021-10-03
Upgrade to Electron 15

### LINKED ISSUE #807: Dictionary plugin started failing due to certificate issues
This started to happen only a couple of days ago.
The dictionary worked fine, but now it shows the following for every query:

image

And in the debug log I'm getting `error: certificate has expired`.

I tried accessing the dictionary API myself with axios and it works fine.
Their certificate is from 2021-09-10, seems to be by LetsEncrypt which swaps certificates every 3 months.

This also happens with the weather plugin, but not the currency or translation plugins.

Is there some sort of certificate caching going on in Electron/Node/Axios?
Is this happening to you too?

Running "clear caches" didn't help, wiping ueli's entire data folder (%APPDATA%\ueli) didn't help either.

NotWearingPants:
Fixes #807 and also brings ueli back to supported Electron versions (the latest!).

I've gone over all of Electron's breaking changes up to version 15, and the only ones affecting ueli are `contextIsolation` which defaults to `true` since v12, and the `electron.remote` module which is deprecated since v12 and removed in v14.

What I did:

- Update `electron` to `^15.1.0`, the latest version
- Added an explicit `contextIsolation: false` to keep the current behavior for now
- ueli specified `enableRemoteModule: true` until now because `electron-store` uses `electron.remote`.
  But the `remote` module was removed in v14, so I've updated `electron-store` to `^8.0.1` (latest) which now uses IPC instead of `remote`.
- `electron-store.Store` is generic on the type of the settings, but `electron-store`s update [changed](https://github.com/sindresorhus/electron-store/commit/b29a67ebec0de843b8c0fbbe1160fe4bc6e540fd) the default from `any` to `unknown`, so I had to specify explicit types for the store in both config & favorites.
- Finally, updating `electron-store` also updated it's dependency `type-fest`.
  `type-fest` had an import of typescript's `esnext` definitions, so ueli's usage of `String.replaceAll` compiled, but `type-fest` removed this import (sindresorhus/type-fest#187) which is good, and so now we must specify we depend on `esnext.string` in `tsconfig.json` for `replaceAll`.


Karan-Munjani (approved):

oliverschwendener:
Thanks for your efforts! I quickly tested it on my Macbook and it works but the startup time is way slower. It takes around 10s for the app to respond when triggering the global shortcut to show the window. Any idea what that could be?

IRlyLikeBac0n:
Hi, any update on this? I really valued the dictionary feature, so wish to see #807 resolved. Seems this is the easiest/best solution to that, with very little potential downside.

oliverschwendener:
No success so far. I disabled all plugins to see if one of the plugins is causing the issue but it's still the same. I measured the time it takes to create the main window and it's arond 15s which is a big downside in my opinion ;) I don't want to merge this PR as long as the long startup time is an issue. Unfortunately currently I have very little time to look into this issue, so let's hope that someone might find a clue.

EdRice4:
I want to introduce myself by first thanking you, @oliverschwendener, as well as everyone else who has contributed to ueli development; it has become an essential application for me! Secondly, everyone's  saying that I am a _novice_ programmer especially with respect to node.js development. However, because I use and enjoy ueli so much and miss the dictionary functionality, I want to contribute in anyway I can.

That being, said, and while this is not much, I want to say that after checking out this PR and testing on my machine, I had no issues with startup time delays: Using the global shortcut, the main window appears just as quickly as it does with the stable version and #807 is resolved.

oliverschwendener:
Somehow I managed to upgrade electron to version 17 without having the startup lag issue on macOS. So i consider this issue as resolved and I will close this PR. Thanks alot again for your efforts and the contribution and sorry that we were not able to merge your work but it's highly appreciated!

PR #817, opened 2021-10-09
Fix links to the old dictionary API in readme/settings

SnirBroshi:
The API change happened in #459 but the links in the README and settings page were never updated


oliverschwendener:
Thanks!

PR #818, opened 2021-10-09
Handle errors of all plugins during index refresh

tkohlmeier:
Currently, when triggering a manual index refresh and one of the plugins throws an error, the `Refreshing indexes...` indication is immediately removed, even if index refreshing of other plugins is still ongoing. Also, only the first error is logged, even if multiple plugins throw errors.

This fixes the issue by using `Promise.allSettled()` instead of `Promise.all()`.


oliverschwendener:
Cool thanks alot!

Issue #574, opened 2020-11-16
Set web search icons by default from url


weiw11:
Given a website url such as  `google.com`, we can grab the favicon by using `https://google.com/favicon.ico` as the icon for the web search.

For every other site:
`https://[website].[domain]`'s favicon can be grabbed with `https://[website].[domain]/favicon.ico`

Majority of the websites follows the same url scheme to store their favicons.

PR #824, opened 2021-10-14
Get favicon from websearch url

weiw11:
Given a domain `https://[website].[domain]`. We can grab the favicon using it's url with `https://[website].[domain]/favicon.ico`.

Alongside this implementation is a fallback to the default icon if the favicon doesn't exist on `icon-component.ts`. 

Based on issue #574 which I created. After the user inputs the url and exits, the Image URL section will auto fill with the `favicon` url. 

Example of change:

### DIFF
--- a/src/renderer/settings/modals/websearch-editing-modal-component.ts
+++ b/src/renderer/settings/modals/websearch-editing-modal-component.ts
@@ -50,6 +50,17 @@ export const websearchEditingModal = Vue.extend({
                 );
             }
         },
+        onUrlBlurGetFaviconUrl() {
+          const websearchEngine: WebSearchEngine = this.websearchEngine;
+          if (websearchEngine.url.length > 0) {
+            try {
+              const url = new URL(websearchEngine.url);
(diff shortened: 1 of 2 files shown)


src/renderer/settings/modals/websearch-editing-modal-component.ts
```diff
@@ -121,7 +132,7 @@ export const websearchEditingModal = Vue.extend({
                             {{ translations.websearchUrl }}
                         </label>
                         <div class="control is-expanded">
-                            <input class="input" type="url" v-model="websearchEngine.url" :placeholder="getUrlPlaceholder()">
+                            <input class="input" type="url" v-model="websearchEngine.url" :placeholder="getUrlPlaceholder()" v-on:blur="onUrlBlurGetFaviconUrl()">
```
oliverschwendener, line 135: Can we use `@blur` instead of `v-on:blur`?

weiw11: Updated to use `@blur` instead.

oliverschwendener:
Cool thanks alot!

PR #829, opened 2021-10-27
add macOS Monterey support

danitt:
Adds MacOS Monterey to list of supported OS versions


oliverschwendener (approved):
Thanks for your contribution! Looks good to me 💪

PR #831, opened 2021-10-29
[Calculator] Handle mathjs evaluating to null

NotWearingPants:
math.js supports the symbol `null`, so `math.evaluate('null') === null`.

Typing `null` in ueli gives an exception, and the log says `Cannot read property 'toString' of null`.
ueli tries to convert the evaluation result to a string, so instead of `result.toString()` I changed it to `String(result)` to support `null`

### DIFF
--- a/src/main/plugins/calculator-plugin/calculator.test.ts
+++ b/src/main/plugins/calculator-plugin/calculator.test.ts
@@ -61,6 +61,15 @@ describe(Calculator.name, () => {
                 expect(actual).toBe(expected);
             });
         });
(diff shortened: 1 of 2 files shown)


oliverschwendener:
Thanks for your contribution! Can we somehow add this case to the unit tests of in `calculator.test.ts` ?

NotWearingPants:
@oliverschwendener I've added a test for `null`

oliverschwendener (approved):
Cool, thank you very much!

Issue #94, opened 2019-01-20
[feature request] For linux


zamgill:
Hi! I love this app. Please create a linux version for this :smile:

jcklpe:
Hey just wanted to put my support in for this feature. I currently work on Windows (personal laptop), linux (home cloud server), and mac (work laptop). I use Listary, Albert, and Alfred for these three and looking at this project it seems to cover most of the features I'd like to use, and if I could use a single app between all 3 I would be extremely happy.

jcklpe:
Just wanted to say that I had to replace my windows laptop so I decided to give Ueli a whirl. It's great! I would love to have this on Linux even in a reduced capacity. I really just use my launcher for search engines (translating, image searching, etc), finding files and folders, running the occasional command, and quick calculations. 

This launcher has great visual design and is super snappy though. It would be awesome to have this on Linux. I realize that some of the more systems oriented stuff like system settings or systems commands might be a little bit harder to make compatibility for, but alfred, which is the albert clone for Linux I currently use, doesn't log any of the system settings stuff, and the shutdown and restart etc commands should be relatively simple to integrate, even if it was on an ad hoc basis by users through running a command shortcut. 

@oliverschwendener The visual design is on point!

jcklpe:
I'm currently using Ueli full time on my mac and windows machines. If anyone is looking for a linux equivalent then I highly recommend [[Albert]](https://github.com/albertlauncher/albert). 

It has a very similar feature set and set-up style. I would really prefer to use Ueli on all my OS or use Albert on all my OS, but for now this is a pretty decent stop gap and there's not a ton of confusion between using the two, though I do prefer how Albert allows websearches with automatic trigger words rather than requiring the "?" mark to trigger a websearch.

jcklpe:
I know this is a big ask but this feature would be great to have. Is there anyway that I could help out with this? I'm not a very good developer but I like fiddling around with my home linux machine. Is it simply a matter of setting up Electron to make a Linux build? Or do other things need to be changed?

oliverschwendener:
I would suggest that you first try to get the app running on a linux machine:

1. Clone the repo
2. Install node 12 and yarn
3. Install the dependencies with `yarn install`
4. Start the watcher (transpiling TS to JS files) with `yarn bundle:watch`
5. Start the app with `yarn start`

Probably it won't work out of the box because it only checks if the app is running on Windows and if not it assumes it is on macOS.

Challenges you might face:
[...]

PR #854, opened 2021-12-28
Feature: Linux Support

ke1v:
Added support for GNOME-based Linux distros (#94) 

Tested on:
- Linux Mint 20.2 Cinnamon
- Ubuntu 20.04 GNOME/Unity

### What I did (built on core components that @yurinnick added)
- Fixed bugs in application searcher
    - Improved application searching and icon generation to be _mostly_ consistent with native desktop
        - Incompatibility between `npm/ini`'s `parse` and Linux `.desktop` files means that some "applications/services" might be shown/hidden when they're not supposed to be
        - App icon generation is not perfect, some icons may differ from native
    - Parsed application names properly

With my limited testing, the rest of the plugins (except bookmark) seem to be working (thanks to @yurinnick)

### What still needs to be done

- Fix bookmark plugin
- Parse `.desktop` files properly 
- Improve native icon search algorithm

### Other notes

- The operating systems plugin is redundant as many Linux system settings are their own application, accessible through the application searcher plugin

I will try to fix the rest of the issues and any more that may arise, but this should be close to ready for GNOME-based Linux desktops. I will try to add support for more desktop environments when I have time.

This is my first contribution/pull request, please let me know if I did anything incorrectly and I'll fix it :D


kristus123:
why was this closed ?

PR #872, opened 2022-02-02
Default Target Currency

peterphmikkelsen:
## Added the option for setting a default target currency.
Hey there! I would like to add the feature "Default Target Currency".
I love using ueli and do so every single day. One of the (very few) things that I thought would be nice to have is the option for setting a default target currency.
### What I did
- Added an option in the settings window under the Currency Converter plugin. The option allows a person to set a default target currency.
- Added the necessary code for checking the input and converting it correctly.

### Example
To see how the feature works, simply leave out the in/to keyword and the target currency:
electron_rwcgnRrN0M

### What I didn't do
- I did not add translations of the name of the option, since I unfortunately don't speak any of the supported languages.

I think it is ready to be merged (except the translation part), but I am of course happy to make any changes you see fit!
Hope you find my small contribution useful.

Thank you!

### DIFF
--- a/src/common/config/currency-converter-options.ts
+++ b/src/common/config/currency-converter-options.ts
@@ -1,9 +1,13 @@
+import { CurrencyCode } from "../../main/plugins/currency-converter-plugin/currency-code";
+
 export interface CurrencyConverterOptions {
     precision: number;
+    defaultTarget: CurrencyCode;
     isEnabled: boolean;
 }
 
 export const defaultCurrencyConverterOptions: CurrencyConverterOptions = {
     isEnabled: false,
     precision: 2,
+    defaultTarget: CurrencyCode.EUR
 };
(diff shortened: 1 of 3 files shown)


GovindKrishnan:
Noobish Doubt : What if there is a file named "12 USD"?

I believe search suggestions don't come up when currency converter is running

oliverschwendener:
> Noobish Doubt : What if there is a file named "12 USD"?
> 
> I believe search suggestions don't come up when currency converter is running

Correct.

GovindKrishnan:
> > Noobish Doubt : What if there is a file named "12 USD"?
> > I believe search suggestions don't come up when currency converter is running
> 
> Yeah, but I guess that would be true for the "12 USD to EUR" case too. Do you consider this a problem? 🙂

It ain't a problem specific to this PR actually. No worries. Just opened another issue #880 to address that. 👍

oliverschwendener (approved):
Thanks for the efforts!

PR #891, opened 2022-03-31
Add /System/Applications to default search path

343max:
Most macOS apps now seem to live in `/System/Applications` so this path should be added to the default setup so users can easily open QuickTime Player and TextEdit again.


oliverschwendener (approved):
Sounds good, thanks for your contribution!

343max:
Thanks for merging so quickly! Amazing too, more contributions will be coming soon.

PR #909, opened 2022-05-08
Settings UI Dark Mode

cjahv:
I added dark mode to the settings, which currently only follows the system.

Using css color inversion with very little code.

Forgive my poor English, use deepl to translate.

### DIFF
--- a/styles/settings.css
+++ b/styles/settings.css
@@ -18,6 +18,9 @@ html {
     .settings button.is-success {
         filter: invert(90%);
     }
+    .settings button.is-danger {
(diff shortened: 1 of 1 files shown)


oliverschwendener:
Thanks for your contribution! Can you please create separate PRs for each feature? That way it would be way easier to review the code.

cjahv:
I deleted irrelevant content

cjahv:
I didn't notice the bulma, I think we can use https://jenil.github.io/bulmaswatch/cyborg/ as the dark theme

PR #910, opened 2022-05-09
Fix ueli command cannot be searched after translation

cjahv:


oliverschwendener:
Am I missing something here? As far as I know it's possible to search en ueli command after changing the language, see screenshot.

image

PR #929, opened 2022-06-21
Setup Crowdin

### LINKED ISSUE #869: The project is missing localization management
Here: Crowdin as an example of Localization Management Tool
Crowdin Integration would be an opportunity to translate ueli into more languages making it accessible etc.
There are Crowdin alternatives if you don't like it.

I've tried adding translation, but it's pain to do inside a text editor - so I gave up finishing it.

andrii-bodnar:
Hey everyone,

Just did some analysis of the current i18n workflow and it feels like GitHub issues and PRs reported by the community are not so easy to handle. The community translators can work together in Crowdin, translations would be delivered as PR and volunteers can raise issues for problematic strings via Crowdin.

I'm suggesting the integration with Crowdin via GitHub Actions. Crowdin is free for open-source projects. This integration works in the following way:
- the action runs every 6 hours (actually, it's up to you what trigger to use, it could also be a push to the default branch, for example)
- upload new source texts to the Crowdin project
- upload existing translations to Crowdin (using the `upload_translations` action config parameter, it's necessary only for the first time)
- download all the new translations from Crowdin and commit these translations to the `localization_branch_name`
- open a Pull Request with the latest translations.

You can find my demo Crowdin project here - [ueli-demo](https://crowdin.com/project/ueli-demo).

Example of the first PR that will be created by Crowdin Action - https://github.com/andrii-bodnar/ueli/pull/1 (Don't worry about the diffs. Crowdin tries to save the initial file structure as much as possible but in the case of the `.ts` files, there is a compromise with the indentations. The next PRs will include new translations only)

Also, I made some improvements to the translation set manager since Crowdin can't replace the const name during file export.

Fixes #869

### DIFF
--- a/.github/workflows/localization.yml
+++ b/.github/workflows/localization.yml
@@ -1,0 +1,40 @@
+# This workflow will run Crowdin Action that will upload new texts to Crowdin, download the newest translations and create a PR
+# For more information see: https://github.com/crowdin/github-action
+
+name: Crowdin Action
+
+# Controls when the action will run.
+on:
+  schedule:
+    - cron: '0 */6 * * *' # Every 6 hours - https://crontab.guru/#0_*/6_*_*_*
+  push: #temporary
+    branches: [ dev ]
+
+jobs:
+  synchronize-with-crowdin:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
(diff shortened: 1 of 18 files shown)


andrii-bodnar:
Hi @oliverschwendener, could you please check this PR? 🙂

PR #988, opened 2023-02-10
lorem-ipsum-plugin

moritz-hoelting:
Adds the option to copy a specified amount of sample text just by typing "lipsum".


README.md
```diff
@@ -130,7 +130,9 @@ This plugin enables you to use the native macOS search to find files and folders
 
 This plugin enables you to browse through your local file system. To start you have to enter a valid absolute filepath.
 
-![file-browser](assets/feature-file-browser.png)
```
oliverschwendener, line 133: Did you remove this image intentionally? If so, why?

src/common/config/lorem-ipsum-options.ts
```diff
@@ -0,0 +1,9 @@
+export interface LoremIpsumOptions {
+    isEnabled: boolean;
+    prefix: string;
+}
+
+export const defaultLoremIpsumOptions: LoremIpsumOptions = {
+    isEnabled: true,
```
oliverschwendener, line 7: As this is a plugin for very specific users, I suggest to make it disabled by default.

.npmrc
```diff
@@ -0,0 +1,3 @@
+node-linker=hoisted
+public-hoist-pattern=*
+shamefully-hoist=true
```
oliverschwendener, line 3: As we're using pnpm for this project, is this needed?

oliverschwendener (changes requested):
Thank you for your contribution! I left some comments.

.npmrc
moritz-hoelting: It was needed for me to run "pnpm package", but I will remove it from the public repository

README.md
moritz-hoelting: No, it was a mistake, thanks for mentioning

src/common/config/lorem-ipsum-options.ts
moritz-hoelting: > Thank you for your contribution! I left some comments.

Everything you commented should be fixed now. Thanks for the fast reply

oliverschwendener (approved):
Nice, it works as expected.

PR #990, opened 2023-02-12
notes-plugin

moritz-hoelting:
This plugins enables to save notes by using prefix "note:" and searching notes by using prefix "notes?"


oliverschwendener:
I am currently rewriting the whole app for v9 (see latest changes in the main branch). Please see if you can port this plugin to an extension in v9.

PR #1006, opened 2023-05-02
Add functionality to queue and run multiple search result item executions

oscardianno:
This PR adds the ability to execute multiple search result items quickly and straight away from a single launcher window.

The motivation for adding this feature is that one might want to quickly launch multiple applications at once, without having to open ueli's window again after each execution. This could be doable by disabling the "_Hide window after execution_" and "_Hide window when focus is lost_" settings. However, the user input didn't clear itself automatically, and sometimes the applications launched would be focused, requiring the user to manually focus the ueli's window again.

So, a queue functionality was implemented by adding a new shortcut key combination "**Alt**" + "Enter", which adds a search result item to the queue instead of executing it immediately. The user input is then cleared automatically. Finally, the queued search result items are executed when the user hides the main window, which can be triggered by pressing the "Esc" key, or, by losing focus or executing a search result item, depending on the app's settings.

Example of functionality:

https://user-images.githubusercontent.com/46467428/235762488-bf206f94-2da5-48ee-be75-ad6e9e8e6e75.mp4

### DIFF
--- a/src/common/ipc-channels.ts
+++ b/src/common/ipc-channels.ts
@@ -2,6 +2,7 @@ export enum IpcChannels {
     search = "search",
     searchResponse = "search-response",
     execute = "execute",
+    emptyExecute = "empty-execute",
     autoComplete = "autocomplete",
     autoCompleteResponse = "autocomplete-response",
     openSearchResultLocation = "open-search-result-location",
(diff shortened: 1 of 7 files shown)


oliverschwendener:
Closing this PR because it is obsolete due to the rewrite for v9. Let's see if we can solve this problem in the new Shortcuts extension. There, a user could define how the shortcut should behave (launch multiple things, and don't hide the window after execution). Sorry for the circumstances.

Issue #1007, opened 2023-05-04
Can't use ueli while terminal is open


kristus123:
i made a workflow that does a simple ping

https://i.gyazo.com/89f3e3eb863062e9ed32dc48af8ddc83.mp4

But ueli is locked until i manually close the terminal

a workaround for this is to append `& exit` at the end to make it close the terminal

kristus123:
btw. ueli is amazing!!!!

kristus123:
@oliverschwendener 

https://github.com/oliverschwendener/ueli/commit/54d57f8e3d94b08cbbe7de43451f625620e38c26

this was a fix that makes it so i can open multiple cmd windows without having it make ueli hang indefinitely

kristus123:
i am no js expert, so i just made it quick and dirty*

kristus123:
i have no idea how i can test this using powershell etc so i can't know if this is a bug with all the other windows terminals

oliverschwendener:
Is this still an issue with the latest beta release of v9? https://github.com/oliverschwendener/ueli/releases
If so, feel free to reopen.

PR #1014, opened 2023-05-30
do not make terminal hang when executing command via cmd

kristus123:


kristus123:
https://github.com/oliverschwendener/ueli/issues/1007

kristus123:
@oliverschwendener Is it possible to get this bug fixed in the next release ? 
And when is the next release ?

oliverschwendener:
I am currently rewriting the whole app for v9. I think this can be solved in the shortcuts extension (that should include workflows in the future). There a user can define if their workflow should wait for the command to finish or not. Sorry for the circumstances.

kristus123:
good to hear, and cool that you are working so actively on v9 :D

kristus123:
Would you say the new code base will be smaller and easier for others to contribute to ?

oliverschwendener:
I think the new code base has better separation. It has "core" modules and "extension" modules. The core modules are responsible for the core functionality and do not know anything about extensions (formerly called plugins). Then the extensions can make use of the core modules and do a very specific thing (e.g. file search, web search, browser bookmarks, whatever). The new code base should also make it much easier to work on an extension because you can do it in isolation and thus code reviews should be smaller. I'm also planning to assign contributors as owners of their extensions, so they can review future changes in the area of their responsibility.

PR #1025, opened 2023-07-24
Some fixes to Russian translations

polyankaglade:
Translated some untranslated sections and fixed some grammar.

Notes:
1. I named "Sleep modes" after how they are named in Windows 11 and macOS, however, there are several options that people use very interchangeably: Сон, Режим сна, Спящий режим. It could be nice to have all of those options available.
2. There is no value for two settings options, so they appear untranslated: 
2.1. General -> "Allow window move" = "Разрешить перемещение окна" 
2.2. Currency Converter -> "Default Currency Target" = "Целевая валюта по умолчанию"

Thank you for such a great app!


oliverschwendener:
I'm just going to accept this blindly as I don't speak Russian 😄

oliverschwendener:
Thanks a lot!

Issue #942, opened 2022-08-09
[Feature Request] Use "Ctrl+ ~"  or "Ctrl+`" as hotkey


makar0nin:
Is it possible to add using a combination with `~` or `backquote`  as hotkeys ?

According to this [link](https://www.electronjs.org/docs/latest/api/accelerator) the use of punctuation is applicable

oliverschwendener:
This should be fixed with the next release.

PR #1028, opened 2023-08-01
Add backtick hotkey

mateuszkozakiewicz:
solves #942


oliverschwendener:
I tested it and I couldn't get it to work. But according to the [electron docs](https://www.electronjs.org/docs/latest/api/accelerator) backticks are supported.

oliverschwendener:
Thanks for your contribution!

mateuszkozakiewicz:
I tested [Alt, Shift, Ctrl] + ` and it is working fine
Super+` works when you disable default quake mode binding in windows
Cmd and Option not tested
AltGr+` does not work

PR #1035, opened 2023-09-10
Update currency converter description

nathanlytang:
Update currency conversion URL in documentation and description to the current provider (https://exchangerate.host)


oliverschwendener (approved):
Thanks for your contribution!

PR #1043, opened 2023-10-01
feat: Use fawazahmed0/currency-api for currency exchange

wywywywy:
So currency exchange is broken right now, because exchangerate.host now requires an API key.

I propose that we migrate to [fawazahmed0/currency-api](https://github.com/fawazahmed0/currency-api) instead, because:

1. It's fast & reliable (hosted on jsdelivr.net)
2. It's free forever
3. Its API is stable
4. It supports a lot more currencies (including crypto)

The risk is that @fawazahmed0 may stop maintaining it one day. But this is in my opinion a lower and less likely risk than a commercial provider start charging their free customers.


oliverschwendener (approved):

oliverschwendener:
Makes sense to me, thanks for your contribution!

fawazahmed0:
@wywywywy Also consider adding [fallback mechanism](https://github.com/fawazahmed0/currency-api/issues/45) in your code:

Here is a JS example for fetching usd to eur rate:

```js
const fetchWithFallback = async (links,obj) => {
  let response;
  for(let link of links)
  {  try{
      response = await fetch(link,obj)
      if(response.ok)
          return response
        }catch(e){}
  }
   return response
}

let from = `usd`
let to = `eur`

fetchWithFallback([
`https://cdn.jsdelivr.net/gh/fawazahmed0/currency-api@1/latest/currencies/${from}/${to}.min.json`,
`https://cdn.jsdelivr.net/gh/fawazahmed0/currency-api@1/latest/currencies/${from}/${to}.json`,
`https://raw.githubusercontent.com/fawazahmed0/currency-api/1/latest/currencies/${from}/${to}.min.json`,
`https://raw.githubusercontent.com/fawazahmed0/currency-api/1/latest/currencies/${from}/${to}.json`
]).then(res => res.json()).then(console.log)
```

PR #1064, opened 2024-01-27
Fix translation plugin

### LINKED ISSUE #999: 404 error when trying to use translation
Translation seems to be broken. I get an 404 error when I try to use it...

tkohlmeier:
This fixes #999 by updating translation plugin to use [v2 linguee api](https://linguee-api.fly.dev/docs#/default/translations_api_v2_translations_get).

Note that this is [not an official API](https://github.com/imankulov/linguee-api/discussions/39) and does not work for all words.

Works fine:
https://linguee-api.fly.dev/api/v2/translations?query=tree&src=en&dst=de

Returns Internal Server Error:
https://linguee-api.fly.dev/api/v2/translations?query=order&src=en&dst=de

So we should find a better API.

### DIFF
--- a/src/main/plugins/translation-plugin/linguee-translator.ts
+++ b/src/main/plugins/translation-plugin/linguee-translator.ts
@@ -1,37 +1,20 @@
 import axios from "axios";
 
-interface WordType {
+interface Translation {
+    text: string;
     pos: string;
 }
 
-interface Translation {
-    text: string;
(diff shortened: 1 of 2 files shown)


oliverschwendener (approved):
Thanks for looking into this. I didn't look through the code thoroughly but I trust you that this works as expected.

PR #1065, opened 2024-01-27
Add ueli command for opening debug log

### LINKED ISSUE #348: Option to open log on error?
Hi, 

If an error occurs a message is displayed instructing the user to check the log. I am proposing to add the ability for the user to instantly open the log file by pressing ENTER.

Users can open the debug log from within the General settings menu but I feel that it is rather hidden away.

If you're happy with this then I can attempt a PR.

Thanks.

image

image

tkohlmeier:
This adds an ueli command for opening the debug log file:

grafik

Additionally this command is now used for the general error search result and thereby fixes #348:

grafik

I only updated German and English versions of the generalErrorDescription, the rest still needs to be updated.

### DIFF
--- a/src/common/error-search-result-item.ts
+++ b/src/common/error-search-result-item.ts
@@ -1,15 +1,16 @@
 import { SearchResultItem } from "./search-result-item";
 import { PluginType } from "../main/plugin-type";
 import { defaultErrorIcon } from "./icon/default-icons";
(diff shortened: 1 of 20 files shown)


oliverschwendener (approved):
Also here, didn't look through the code as I trust you. Thanks a lot for looking into this!

PR #1066, opened 2024-01-28
Prevent main window from being out of bounds

### LINKED ISSUE #871: Putting to high of a value in
If you put a high value (for me I put 50) into the max search results per page. Then hit "Alt+Space" to open ueli, ueli no longer appeared on screen.

It seems that as you increase this number ueli opens higher on the screen. What I think is happening is it is appearing out of bounds of the current screen and thus not visible.

image

tkohlmeier:
This fixes #871 by ensuring main window x/y location is at least 0.


oliverschwendener (approved):
Nice! Thanks alot

PR #1071, opened 2024-03-01
Fix currency-api link

fawazahmed0:


wywywywy:
Is it worth adding the Cloudflare URL in as fallback as well?

oliverschwendener:
What's the issue? Can you add a more detailed description?

oliverschwendener (approved):

tanihot1987 (commented):
Gut

PR #1076, opened 2024-03-04
test: check if prettier check fails

oliverschwendener:
test


oliverschwendener:
Nice, prettier fails as expected.

PR #1077, opened 2024-03-08
feat(IniFileParser): added IniFileParser

oliverschwendener:
This PR adds a utility for parsing ini file content. This is needed for adding Linux support for the ApplicationSearch extension.


ke1v:
I'd advise against using the `ini` npm package as it doesn't play well with Linux ini files, since there is no official ini standard. It's why I wrote a custom ini parser for Linux support

ke1v:
The biggest issue with `ini` is comments, since [Linux uses](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#comments) `#` for comments while `ini` uses `;`. This lead to a couple issues when a Linux .ini file would have a semicolon separated list and `ini` parses it as a comment

oliverschwendener:
Cool, thanks for your advise. I'll revert it then and I will add your implementation.

oliverschwendener:
@ke1v If you're happy with the current implementation, I'll merge this later, so we can use it in your PR.

oliverschwendener:
@ke1v Would you like to take over the ownership of this module as you seem to have more knowledge around the `.ini` topic? :) This means that any future change in the `src\main\Core\IniFileParser` folder needs to be reviewed by you and you can change the internals of this module freely without review from me or anyone else. Let me know if you're interested.

ke1v:
@oliverschwendener I've made some improvements to the parser (#1079) and it should be good to merge. Also, I'd be happy to take over reviews for this module :)

PR #1073, opened 2024-03-03
Feature: Linux support for Application Search

ke1v:
Ported over the previous attempt to add Linux support into the v9 codebase

Currently it's broken due to the usage of `extract-file-icon`, which doesn't support Linux. I haven't modified FileImageGenerator since it is a core dependency. How should I approach integrating Linux support into the application?

PR #1078, opened 2024-03-08
feat(Application Search): Linux Support

ke1v:
Reopening #1073


ke1v:
Progress update - Extension is functional, but is still:
- Missing flatpak apps
- Packaging for Linux platforms fail. Running the either packages give the runtime error `error: Cannot find module '@babel/runtime/helpers/interopRequireDefault'` or doesn't render
- Need to write tests for new functionality


Working Screenshot (ran using `pnpm dev`)j:
image

Attempt to package (ran executable from `pnpm package`):
image

oliverschwendener:
Thanks for the update and your efforts, i'll have a look when i'm back from vacation. The failing build is expected, you have to install the pnpm dependencies under the correct configuration (see release.yml where it says "configure pnpm").

oliverschwendener:
To fix the `Cannot find module '@babel/runtime/helpers/interopRequireDefault'` error, you have to run these commands before running `pnpm package`:


```
pnpm config set shamefully-hoist true --location=project
pnpm config set auto-install-peers true --location=project
pnpm config set strict-peer-dependencies true --location=project
```

oliverschwendener:
Can you update your branch, so that the changes that already have been added in my main branch disappear from this PR?

ke1v:
Apologies for the messy commits, I'm still learning git. The plugin seems to be fully functional in GNOME and should be good to merge, but I don't have time currently to test out the other popular desktops for Linux.

src/main/Core/FileSystemUtility/Contract/FileSystemUtility.ts
```diff
@@ -4,6 +4,9 @@ export interface FileSystemUtility {
     pathExists(fileOrFolderPath: string): Promise<boolean>;
     readJsonFile<T>(filePath: string): Promise<T>;
     readJsonFileSync<T>(filePath: string): T;
+    readFile(filePath: string): Promise<Buffer>;
+    readFileSync(filePath: string): Buffer;
+    copyFile(srcPath: string, destPath: string): Promise<void>;
```
oliverschwendener, line 9: `copyFile` is already specified on the bottom, so no need to add this.

src/main/Core/ImageGenerator/LinuxIconExtractor.ts
```diff
@@ -0,0 +1,354 @@
...
+            [join(homePath, "Music")]: "folder-music",
+            [join(homePath, "Pictures")]: "folder-pictures",
+            [join(homePath, "Public")]: "folder-publicshare",
+            [join(homePath, "Templates")]: "folder-templates",
+            [join(homePath, "Bookmarks")]: "user-bookmarks",
+        };
+    }
+
+    public matchesFilePath(filePath: string) {
+        // TODO: Add AppImage support
```
oliverschwendener, line 70: Please remove this TODO comment. Either this todo needs to be addressed before merging or it should be removed.

src/main/Core/ImageGenerator/LinuxIconExtractor.ts
```diff
@@ -0,0 +1,354 @@
...
+import type { Logger } from "@Core/Logger";
+import type { Image } from "@common/Core/Image";
+import { dirname, extname, join } from "path";
+import type { CacheFileNameGenerator } from "./CacheFileNameGenerator";
+import type { FileIconExtractor } from "./FileIconExtractor";
+
+// Specs from https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html
+interface IconThemeSubdir {
+    Size: number;
+
```
oliverschwendener, line 13: Is there a reason for this empty line? If not, it should be removed.

src/main/Core/ImageGenerator/LinuxIconExtractor.ts
```diff
@@ -0,0 +1,354 @@
...
+        }
+        return values;
+    }
+
+    private async extractAppIconName(filePath: string) {
+        const appConfig = this.iniFileParser.parseIniFileContent(
+            (await this.fileSystemUtility.readFile(filePath)).toString(),
+        )["Desktop Entry"];
+
+        if (!appConfig["Icon"]) throw `${filePath} doesn't have an icon!`;
```
oliverschwendener, line 104: Please remove all exclamation marks from the error messages because they add no value.

```suggestion
        if (!appConfig["Icon"]) throw `${filePath} doesn't have an icon`;
```

oliverschwendener: To match this project's code style I suggest avoiding one-line if-statements. In my opinion it makes the code a bit easier to read.

```suggestion
        if (!appConfig["Icon"]) {
            throw `${filePath} doesn't have an icon!`;
        }
```

src/main/Core/ImageGenerator/LinuxIconExtractor.ts
```diff
@@ -0,0 +1,354 @@
...
+        return `${join(this.cacheFolder, this.cacheFileNameGenerator.generateCacheFileName(iconName))}.png`;
+    }
+
+    private async generateAppIcon(iconName: string, iconFilePath: string): Promise<void> {
+        if (!this.searchCache || !this.userTheme) await this.generateThemeSearchCache();
+
+        const iconSize = 128;
+
+        const iconPath = this.findIcon(iconName, iconSize, 1, this.userTheme);
+        if (!iconPath) throw `Icon ${iconName} could not be found!`;
```
oliverschwendener, line 134: Also here:

```suggestion
        if (!iconPath) {
            throw `Icon ${iconName} could not be found!`;
        }
```

src/main/Core/ImageGenerator/LinuxIconExtractor.ts
```diff
@@ -0,0 +1,354 @@
...
+import type { FileSystemUtility } from "@Core/FileSystemUtility";
+import type { IniFileParser } from "@Core/IniFileParser/IniFileParser";
+import type { Logger } from "@Core/Logger";
+import type { Image } from "@common/Core/Image";
+import { dirname, extname, join } from "path";
+import type { CacheFileNameGenerator } from "./CacheFileNameGenerator";
+import type { FileIconExtractor } from "./FileIconExtractor";
+
+// Specs from https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html
+interface IconThemeSubdir {
```
oliverschwendener, line 11: Please use types instead of Interfaces:

```suggestion
type IconThemeSubdir = {
```

src/main/Core/ImageGenerator/LinuxIconExtractor.ts
```diff
@@ -0,0 +1,354 @@
...
+    Size: number;
+
+    Scale: number; // Defaults to 1
+    MinSize: number; // Defaults to size
+    MaxSize: number; // Defaults to size
+    Threshold: number; // Defaults to 2
+    Type: "Fixed" | "Scalable" | "Threshold"; // Defaults to Threshold
+}
+
+interface IconTheme {
```
oliverschwendener, line 21: Same here

src/main/Core/ImageGenerator/LinuxIconExtractor.ts
```diff
@@ -0,0 +1,354 @@
...
+
+    private lookupFallbackIcon(iconName: string): string {
+        for (const directory of this.baseDirectories) {
+            for (const extension of ["png", "svg", "xpm"]) {
+                const fileName = `${join(directory, iconName)}.${extension}`;
+                if (this.fileSystemUtility.existsSync(fileName)) return fileName;
+            }
+        }
+        if (this.fileSystemUtility.existsSync(iconName)) return iconName;
+        return "";
```
oliverschwendener, line 321: I suggest avoiding returning an empty string. Why not return `undefined` instead? In my opinion that would make it much more clear that no icon could be found/generated.

src/renderer/Extensions/ApplicationSearch/Linux/LinuxSettings.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <SectionList>
+            <Section>
+                <Field
+                    label="Application Folders (Hardcoded - Cannot be changed)"
+                    style={{ display: "flex", flexDirection: "column", gap: 5 }}
+                >
+                    {folders.map((folder) => (
+                        <Input key={`linuxFolder-${folder}`} value={folder} readOnly={true} disabled={true} />
+                    ))}
+                </Field>
```
oliverschwendener, line 24: If the folders are hard-coded, I suggest not showing them in the settings. Looks like we don't need any settings for the ApplicationSearch on Linux at all. I would remove this.

src/main/Extensions/ApplicationSearch/Linux/LinuxApplicationRepository.ts
```diff
@@ -0,0 +1,86 @@
...
+                    }
+                });
+
+            const results = (await Promise.allSettled(promises))
+                .map((result) => (result.status === "fulfilled" ? result.value : null))
+                .filter((result) => result);
+
+            return results;
+        } catch (error) {
+            this.logger.error(`Error in ApplicationSearch! Error: ${error}`);
```
oliverschwendener, line 77: Also here, no need for the exclamation mark. Also I would improve this error message that it says what failed. Here is a suggestion:

```typescript
this.logger.error(`Failed to get linux applications. Reason: ${error}`);
```

src/main/Extensions/ApplicationSearch/Linux/LinuxApplicationRepository.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    public async getApplications(): Promise<Application[]> {
+        try {
+            // For future support of other formats
+            const linuxFolders = this.settings.getValue<string[]>("linuxFolders");
+            const appExtensions = this.settings.getValue<string[]>("linuxFileExtensions");
+
+            this.logger.info(
+                `find -L ${linuxFolders.join(" ")} -type f -name "*${appExtensions.join('" -o -name "')}"`,
+            );
```
oliverschwendener, line 32: Can you explain why we log this?

oliverschwendener:
I tested it in on Ubuntu Desktop (downloaded from here: https://ubuntu.com/download/desktop) and it didn't work because `Desktop environment not supported`. If the system does not support the user's desktop environment, then the extension should not be available, and should not be invoked. Each extension has a method `isSupported`, I suggest to check the desktop environment there and return false if the desktop environment is not supported. Additionally, there are two architectural issues that I would like to address:

- Please remove the `createGtkLaunchAction` method from the `SearchResultItemActionUtility` as this class should not know about extension specific, or platform specific things. I suggest creating a separate class or just a function, whatever you see fit, in the ApplicationSearch extension folder.

- Please move the `GtkLaunchActionHandler` also to the extension folder and specify the action handler in the `bootstrap` method of the `ApplicationSearchModule` like this:
```typescript
export class ApplicationSearchModule implements ExtensionModule {
    public bootstrap(dependencyRegistry: DependencyRegistry<Dependencies>): ExtensionBootstrapResult {
        ...

        return {
            actionHandlers: [
                new GtkLaunchActionHandler(...), // <= here
            ],
            extension: new ApplicationSearch(...),
        };
    }
}
```

oliverschwendener:
FYI: I addressed all my feedback in this branch.

oliverschwendener:
I tested it on Ubuntu Cinnamon, but none of the applications worked as the icons couldn't be found/generated. I suggest to fall back to a generic app icon when the icon generation fails.

ke1v:
I've also experienced this issue and am working on a fix to this and other bugs. I'm also planning to add support for more desktop environments than Cinnamon and GNOME, should I create a type similar to OperatingSystems with all supported Linux environments?

oliverschwendener:
Yes you can, but let's keep in in the ApplicationSearch folder for now, as we only need it there now.

src/main/Core/ImageGenerator/LinuxIconExtractor.ts
```diff
@@ -0,0 +1,412 @@
...
+        private readonly commandlineUtility: CommandlineUtility,
+        private readonly iniFileParser: IniFileParser,
+        private readonly logger: Logger,
+        private readonly cacheFileNameGenerator: CacheFileNameGenerator,
+        private readonly cacheFolder: string,
+        private readonly homePath: string,
+    ) {
+        this.baseDirectories = [
+            join(this.homePath, ".icons"),
+            join(process.env["XDG_DATA_HOME"] || join(this.homePath, ".local", "share"), "icons"),
```
oliverschwendener, line 47: By the way, we can now use the new `EnvironmentVariableProvider` here instead of reyling on `process.env`.  This should make it easier to mock it in unit tests.

src/main/Core/ImageGenerator/LinuxIconExtractor.ts
```diff
@@ -0,0 +1,412 @@
...
+    name: string;
+    path: string;
+    // Paths in order as listed in index.theme
+    subdirectories: string[];
+    // Map<subdirPath, subdirData>
+    subdirData: Map<string, IconThemeSubdir>;
+    parents: string[];
+};
+
+export class LinuxIconExtractor implements FileIconExtractor {
```
oliverschwendener, line 30: And is there a way to reduce the size of this class? It's already quite big, which is an indicator that it does a lot of different things. This makes it really hard to write maintainable tests. I suggest finding a way to extract some code of this class into other classes or utilities (or just functions).

ke1v: I can try to pull some of the code into a second class, but the reason the class is so large is because finding icons on Linux is a fairly complex process due to the ability to theme their own desktops

ke1v:
I've changed the way applications get launched and it now supports launching files located anywhere on GNOME. I still have other desktops to test out (sorry spinning up VMs take a lot of time) which are all currently listed in `LaunchDesktopFileActionHandler.ts`

oliverschwendener:
I tested it on Ubuntu Cinnamon and it works fine, although some icons couldn't be generated. There is a lot of opportunity for improvements, for example splitting code into multiple classes, simplifying code flows or adding more tests, but this is good enough for now, so let's merge it to move forward. Thanks alot for your time and all your efforts, much appreciated! 💪

For future changes please keep in mind:
- Make sure you set up [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), as it will be required for future contributions
- Add unit tests for every class or function from the beginning. It's often times harder to add them later on. I also had to learn this lesson 😅 
- Use `import type ...` whenever possible
- Use accurate variable names, for example:
  ```typescript
  // instead of this
  const processEnv: EnvironmentVariableProvider = ...
  ```
  ```typescript
  // do this
  const envrionmentVariableProvider: EnvironmentVariableProvider ...
  ```
- Try to keep your code tidy and organized :) In this PR I added a lot of line breaks to make code blocks more separated from each other, so it's easier to read. This is a good example: https://github.com/oliverschwendener/ueli/pull/1078/commits/4d7dbaae4df92561dd32b0b733d8e6456414e7ba

I added you to the code owners of all the Linux parts of the ApplicationSearch extension, so I will forward all future feature request, issues and contributions to you for review, if that's fine. Keep it!

oliverschwendener (approved):

PR #1080, opened 2024-03-11
fix(IniFileParser): Updated Interface

ke1v:


src/main/Core/IniFileParser/Contract/IniFileParser.ts
```diff
@@ -4,14 +4,16 @@
 export interface IniFileParser {
     /**
      * Parses the given ini file content and returns the result as an object.
-     * @param fileString The ini file content, e.g.:
+     * @param {string} fileString The ini file content, e.g.:
```
oliverschwendener, line 7: I would remove the types from these coments, as the types are already defined via TypeScript.

oliverschwendener (changes requested):
See my comment above.

ke1v:
Removed. Sorry it's my first time using JSDocs

oliverschwendener (approved):

PR #1105, opened 2024-05-26
Fix encoding issues in application search

tkohlmeier:
Same topic as #1096 but it now should work correctly by using correct encoding everywhere.

grafik

### DIFF
--- a/src/main/Core/FileSystemUtility/NodeJsFileSystemUtility.ts
+++ b/src/main/Core/FileSystemUtility/NodeJsFileSystemUtility.ts
@@ -16,6 +16,8 @@ import { join } from "path";
 import type { FileSystemUtility } from "./Contract";
 
 export class NodeJsFileSystemUtility implements FileSystemUtility {
+    private byteOrderMark: string = "\ufeff";
+
     public async clearFolder(folderPath: string): Promise<void> {
         const filePaths = await this.readDirectory(folderPath);
         await Promise.all(filePaths.map((filePath) => this.removeFile(filePath)));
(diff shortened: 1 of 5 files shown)


src/main/Core/FileSystemUtility/NodeJsFileSystemUtility.ts
```diff
@@ -60,7 +62,7 @@ export class NodeJsFileSystemUtility implements FileSystemUtility {
 
     public writeTextFile(data: string, filePath: string): Promise<void> {
         return new Promise((resolve, reject) => {
-            writeFile(filePath, data, "utf-8", (error) => {
+            writeFile(filePath, `${this.byteOrderMark}${data}`, "utf-8", (error) => {
```
oliverschwendener, line 65: Can you explain what is happening here? 😅 Are you sure that this won't cause any issues on macOS and Linux?

tkohlmeier: Powershell does not seem to recognize a file as UTF-8 if the [byte-order mark (BOM)](https://en.wikipedia.org/wiki/Byte_order_mark) is missing. Node does not seem to be able to write the BOM itself, so you have to add it manually. Source: https://stackoverflow.com/a/13859239/1836528

I did not test this on macOS or Linux since I don't have those OSes. I could set up a VM to test it on Linux but I don't have a Mac. Can you test it? Or we could just make it OS-dependent.

oliverschwendener: No worries, I can test this on a Macbook next week. If this is an issue on macOS or Linux we can create OS-specific file system utility classes.

oliverschwendener:
Thanks for you fix, I ended up implementing it in the `PowershellUtility` instead of the `FileSystemUtility` as it reduces the vector of the change. See 01c6f6bcd9f8c9fbf5568d52b5573049e2009a1b

PR #1145, opened 2024-07-24
feat: base64 extension

ChristopherSteiner:
Currently I have to encode and decode some strings with base64 on a daily basis, so I thought a simple extension for Ueli would be nice.
Let me know if there are any guidelines on a preferred pattern to trigger an extensions "instant search", I was unable to find anything. Currently this extensions search gets triggered when the user entered three space seperated strings with the following format: `<some string> <b/b64/base64> <encode/decode>`

image
image


oliverschwendener:
I think it would make sense to create a custom UI for this, similar to the DeepL translator, with 2 bigger text fields for encoding and decoding. The generic search field is too small when you want to encode/decode bigger strings. What do you think? I can also help with that if you want.

image

ChristopherSteiner:
Yes this sounds reasonable. I hadn't used this extension before and therefore didn't realise that you can make your own custom UIs. I will take a look at the Deepl implementation and try to adapt it to my use case as soon as I have the time. I'll let you know if I need assistance, but I think it should be pretty straight forward because the UI is more or less the same (or even easier because of the missing api key UI). Even for a TypeScript Newbie like me :)

ChristopherSteiner:
I've removed the instant search results and implemented a new UI for this extension instead.
Basically you can enter a string on either side and the string gets encoded or decoded on the other side.

image

src/main/Extensions/Base64Conversion/Types.ts
```diff
@@ -0,0 +1,9 @@
+export type InvocationArgument = {
+    invocationString: string;
+    encodingType: EncodingType;
+};
+
+export enum EncodingType {
+    Encode,
+    Decode,
+}
```
oliverschwendener, line 9: I generally suggest avoiding "type files" where you store all your types.

Please create a dedicated file for each type. Also I suggest avoiding enums, just use a strings:

```typescript
export type EncodingType = "encode" | "decode";
```

src/main/Extensions/Base64Conversion/Types.ts
```diff
@@ -0,0 +1,9 @@
+export type InvocationArgument = {
+    invocationString: string;
+    encodingType: EncodingType;
+};
```
oliverschwendener, line 4: I find the names "invocationString" and "encodingType" not very good. "decode" is not an encoding type. To me "decode" and "encode" are more like "actions" or something similar. And the "invocationString" is maybe a "payload"? What do you think about this?


```typescript
export type InvocationArgument = {
    action: "encode" | "decode";
    payload: string;
}
```

ChristopherSteiner: I agree, naming is hard and I wasn't paying enough attention when I wrote this type.

src/main/Extensions/Base64Conversion/Types.ts
ChristopherSteiner: I was inspired by the types file from the Deepl Translator [Type](https://github.com/oliverschwendener/ueli/blob/main/src/main/Extensions/DeeplTranslator/Types.ts). 
Out of curiosity why strings over enums?
I've updated the code according to your suggestions, thank you for taking a look.

src/main/Extensions/Base64Conversion/Base64Conversion.ts
```diff
@@ -0,0 +1,87 @@
...
+            return Buffer.from(argument.payload, "base64").toString("binary");
+        }
+    }
+
+    public isSupported(): boolean {
+        return true;
+    }
+
+    public getSettingDefaultValue<T>(key: string): T {
+        return this.defaultSettings[key];
```
oliverschwendener, line 56: As your extension does not have any settings, and therefore also no default settings, you can just return undefined here:

```typescript
public getSettingDefaultValue() {
    return undefined;
}
```

And then you can also remove the private property `defaultSettings`.

src/main/Extensions/Base64Conversion/Base64Conversion.ts
```diff
@@ -0,0 +1,87 @@
...
+        ];
+    }
+
+    public async invoke(argument: InvocationArgument): Promise<string> {
+        if (argument.action === "encode") {
+            return Buffer.from(argument.payload, "binary").toString("base64");
+        } else {
+            return Buffer.from(argument.payload, "base64").toString("binary");
+        }
+    }
```
oliverschwendener, line 49: This is just my personal preference, but I like to do the following:

- I like to destructure objects in function/method arguments
- If I encounter two return statements in an `if/else` then I like to turn it to a single return statement

Here is an example:

```typescript
public async invoke({ action, payload }: InvocationArgument): Promise<string> {
    return action === "encode"
        ? Buffer.from(payload, "binary").toString("base64")
        : Buffer.from(payload, "base64").toString("binary");
}
```

In my opinion the code looks more pleasing and elegant to the eye and is therefore nicer to read.

src/main/Extensions/Base64Conversion/Base64Conversion.ts
```diff
@@ -0,0 +1,87 @@
+import { AssetPathResolver } from "@Core/AssetPathResolver";
+import { Extension } from "@Core/Extension";
+import { Translator } from "@Core/Translator";
+import { SearchResultItem, SearchResultItemActionUtility } from "@common/Core";
+import { Image } from "@common/Core/Image";
+import { Buffer } from "buffer";
+import { InvocationArgument } from "./InvocationArgument";
```
oliverschwendener, line 7: Please import types only when possible:

```suggestion
import type { AssetPathResolver } from "@Core/AssetPathResolver";
import type { Extension } from "@Core/Extension";
import type { Translator } from "@Core/Translator";
import { type SearchResultItem, SearchResultItemActionUtility } from "@common/Core";
import type { Image } from "@common/Core/Image";
import { Buffer } from "buffer";
import type { InvocationArgument } from "./InvocationArgument";
```

src/renderer/Extensions/Base64Conversion/Converter.tsx
```diff
@@ -0,0 +1,76 @@
+import { useContextBridge } from "@Core/Hooks";
+import { Textarea } from "@fluentui/react-components";
+import { useEffect, useState } from "react";
+import { InvocationArgument } from "../../../main/Extensions/Base64Conversion/InvocationArgument";
```
oliverschwendener, line 4: Don't import files across the "main" and "renderer" folders.

The rules are:
- A file in the "renderer" directory is only allowed to import files from "renderer" and "common".
- A file in the "main" directory is only allowed to import files from "main" and "common".

I suggest moving the shared code to the common folder then you can import them like this:

```typescript
import { MyStuff } from "@Common/Extensions/MyExtension/MyStuff";
```

Sorry, you couldn't know these things as they're not documented (yet).

oliverschwendener:
It works as expected, thanks for putting your time into this project! I left some comments, apart from that it looks good to me.

By the way, are you willing to mainain this extension? This means that I'll make you the code owner of all files in the following folders:
- `main/Extensions/Base64Conversion`
- `renderer/Extensions/Base64Conversion`
- `common/Extensions/Base64Conversion`

And I'll forward all issues/questions/feature requests regarding this extension to you.

oliverschwendener:
And what do you think about adding some unit tests? ;)

ChristopherSteiner:
First of all thanks for the thorough code review and explanations. 
I think I was able to refactor everything according to your recommendations. I was actually thinking about unit tests but then I dismissed it because the effective logic of the extensions are like two lines of code :). However I moved the encode and decode methods into there own class, I feel it's nicer like this anyway and I could test them now too. 

And sure, I wouldn't mind to further maintain this extension. If I'm the owner of that code there are hopefully still the code reviews from you/another owner in the PR though?

src/main/Extensions/Base64Conversion/Base64Converter.test.ts
```diff
@@ -0,0 +1,16 @@
...
+            expect(Base64Converter.encode("Test:1886")).toBe("VGVzdDoxODg2");
+        });
+    });
+
+    describe(Base64Converter.decode, () => {
+        it("should return the correct decoded value of a base64 string", () => {
+            expect(Base64Converter.decode("VGhpcyBpcyBhIHRlc3Qh")).toBe("This is a test!");
+        });
+    });
+});
```
oliverschwendener, line 16: Very nice! 💪

src/main/Extensions/Base64Conversion/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./Base64Conversion";
```
oliverschwendener, line 1: Please export the `Base64ConversionModule` only here, as it's the only thing that is needed outside of this folder.

src/main/Extensions/ExtensionLoader.ts
```diff
@@ -2,6 +2,7 @@ import type { Dependencies } from "@Core/Dependencies";
 import type { DependencyRegistry } from "@Core/DependencyRegistry";
 import { AppearanceSwitcherModule } from "./AppearanceSwitcher";
 import { ApplicationSearchModule } from "./ApplicationSearch";
+import { Base64ConversionModule } from "./Base64Conversion/Base64ConversionModule";
```
oliverschwendener, line 5: ...then you can also simplify the import here:

```suggestion
import { Base64ConversionModule } from "./Base64Conversion";
```

src/main/Extensions/Base64Conversion/Base64Conversion.ts
```diff
@@ -0,0 +1,81 @@
+import type { AssetPathResolver } from "@Core/AssetPathResolver";
+import type { Extension } from "@Core/Extension";
+import type { Translator } from "@Core/Translator";
+import { type SearchResultItem, SearchResultItemActionUtility } from "@common/Core";
+import type { Image } from "@common/Core/Image";
+import type { InvocationArgument } from "@common/Extensions/Base64Conversion/InvocationArgument";
```
oliverschwendener, line 6: You can simplify this import as it's exported in `common/Extensions/Base64Conversion/index.ts`:

```suggestion
import type { InvocationArgument } from "@common/Extensions/Base64Conversion";
```

src/renderer/Extensions/Base64Conversion/Converter.tsx
```diff
@@ -0,0 +1,76 @@
+import type { InvocationArgument } from "@common/Extensions/Base64Conversion/InvocationArgument";
```
oliverschwendener, line 1: ...same here:

```suggestion
import type { InvocationArgument } from "@common/Extensions/Base64Conversion";
```

oliverschwendener:
Sorry, for more nit-picky comments, but after this it should be good to merge! Thanks again for your patience :)

oliverschwendener:
> If I'm the owner of that code there are hopefully still the code reviews from you/another owner in the PR though?

If I feel the need, or if you want to get help, sure I can always help out. Otherwise, I will just review the code to make sure that new code does not violate some architectural rules (which, as I said before, are not documented yet). The goal is to enable devs to maintain their extensions as independently as possible, as long as the code meets a certain standard.

src/main/Extensions/Base64Conversion/Types.ts
oliverschwendener: I prefer string literal types over enums for the following reason, let's make a quick example:

### Example 1: Enum

Imagine you have this function definition that takes an enum as an argument:

```typescript
export enum Action {
    encode,
    decode,
}
```

```typescript
import { Action } from "./Action";

export const invoke(action: Action) => {
  // something happens here
};
```

If you want to invoke this function from somewhere else you also have to import the enum type, for example like this:

```typescript
import { Action } from "./Action";
import { invoke } from "./invoke";

invoke(Action.encode);
invoke(Action.decode);
```

In order to pass the right argument to the function you have to import an additional type.

### Example 2: String literal type

If you implement the same function with a string literal type, the additional type import is not needed:

```typescript
export const invoke(action: "encode" | "decode") => {
  // something happens here
};
```

```typescript
import { invoke } from "./invoke";

invoke("encode");
invoke("decode");
```

This is less code and also easier to use in my opinion.

Additionally, the electron API is also full of string literal types instead of enums and I like to stay consistent:

For example:

```typescript
browserWindow.setBackgroundMaterial("mica");

// instead of:
// browserWindow.setBackgroundMaterial(BackgroundMaterial.Mica)
```

But I guess this is because they want to be compatible with vanilla JavaScript which (I think) doesn't have enums.

I hope this helps :)

ChristopherSteiner:
Guess I should read up about the exports and imports :)

oliverschwendener (approved):

oliverschwendener:
Looks good now! Thanks for your contribution.

PR #1147, opened 2024-07-28
fix: Use "kl" instead of "lk" as DuckDuckGo parameter

ykhrito:
DuckDuckGo's region parameter is supposed to be “kl”. 🧐
cf: https://duckduckgo.com/duckduckgo-help-pages/settings/params/


oliverschwendener:
True, thanks alot!

oliverschwendener (approved):

ykhrito:
Sorry, I signed the commit and pushed it again! 🙇

PR #1159, opened 2024-08-12
fix(ApplicationSearch): Add support for more desktop environments

### LINKED ISSUE #1154: [Bug] Failed to generate a cache for icon extractor
**Describe the bug**
I installed the Appimage on Majaro and after I started the App I see only the Ueli commands.
The following Error appears in the logs:
`[8.8.2024, 20:39:34][ERROR] Failed to generate a cache for icon extractor. Reason: TypeError: {(intermediate value)(intermediate value)}[e] is not a function`

**To Reproduce**
1. Install Manjaro
1. Install the App via Appimage
1. Start the App
2. check the logs

**Expected behavior**
Search result of Apps and later for files and other stuff

**Screenshots**

**Environment**
 - OS: Latest Manjaro Linux 6.6.32-1-MANJARO 1 SMP PREEMPT_DYNAMIC Mon May 27 03:41:25 UTC 2024 x86_64 GNU/Linu

**Additional context**

ke1v:
Fixes #1154, adds support for KDE, MATE, XFCE, and Pantheon desktops

Also includes fix for small bug in `IniFileParser`


src/main/Core/ImageGenerator/Linux/LinuxAppIconExtractor.ts
```diff
@@ -271,15 +273,52 @@ export class LinuxAppIconExtractor implements FileIconExtractor {
     }
 
     private async getIconThemeName(): Promise<string> {
-        const currentDesktopEnvironment = this.environmentVariablePRovider.get("XDG_SESSION_DESKTOP");
-
-        const iconThemeNameMap: Record<"cinnamon" | "gnome", () => Promise<string>> = {
-            cinnamon: () =>
-                this.commandlineUtility.executeCommand("gsettings get org.cinnamon.desktop.interface icon-theme"),
-            gnome: () => this.commandlineUtility.executeCommand("gsettings get org.gnome.desktop.interface icon-theme"),
+        const iconThemeNameMap: Record<LinuxDesktopEnvironment, () => Promise<string>> = {
```
oliverschwendener, line 276: I suggest using just strings here to untangle the `LinuxAppIconGenerator` from the application search extension.

```suggestion
        const iconThemeNameMap: Record<string, () => Promise<string>> = {
```

src/main/Core/ImageGenerator/Linux/LinuxAppIconExtractor.ts
```diff
@@ -271,15 +273,52 @@ export class LinuxAppIconExtractor implements FileIconExtractor {
...
+            GNOME: (): Promise<string> => {
```
oliverschwendener, line 277: I think these type declarations are not necessary:

```suggestion
            GNOME: () => {
```

Also, you can simplify this by removing the redudant return statement:

```typescript
GNOME: () => this.commandlineUtility.executeCommand("gsettings get org.gnome.desktop.interface icon-theme")
```

src/main/Core/IniFileParser/IniFileParser.ts
```diff
@@ -26,19 +26,20 @@ export class IniFileParser implements IniFileParserInterface {
         for (const line of fileLines) {
             // New group
             if (line.startsWith("[") && line.endsWith("]") && !line.slice(1, -1).match(/\[\]/m)) {
-                if (result[line]) {
-                    throw `Ini Format Error! Duplicate groups found in ${fileString}`;
+                group = line.slice(1, -1);
+
+                if (result[group]) {
+                    throw new Error(`Ini Format Error. Duplicate group ${group} found`);
```
oliverschwendener, line 32: Is this related to the application search extension? If not, please create a separate PR for it.

src/common/Core/LinuxDesktopEnvironment.ts
```diff
@@ -0,0 +1,16 @@
...
+    "GNOME",
+    "GNOME-Classic",
+    "GNOME-Flashback",
+    "KDE",
+    "MATE",
+    "XFCE",
+    "Cinnamon",
+    "X-Cinnamon",
+    "Pantheon",
+] as const;
```
oliverschwendener, line 14: In my opinion this list of supported linux desktop environments should not be in `common/Core` one extension or module could support a few of those and another module could support others. I think it's wrong to have a global list of supported environments.

I suggest moving this to the application search extension and to the file image generator module.

src/main/Core/IniFileParser/IniFileParser.ts
ke1v: Just a bug I caught while testing the app search extension. I'll make a separate PR for it

src/common/Core/LinuxDesktopEnvironment.ts
ke1v: I had completely overlooked that modules might support different environments. I will move checking support to the individual modules. Could I keep `LinuxDesktopEnvironmentModule` as well as a list of all Linux environments in `common/Core`, supported or not, to have a common type to compare Linux desktops?

oliverschwendener: If it's just a type I'm fine with keeping it. I suggest just using a string literal type:

```typescript
export type LinuxDesktopEnvironment = "GNOME" | "GNOME-Classic" ...
```

oliverschwendener: I added a suggestion here: #1172

ke1v:
@oliverschwendener Everything should be in working order now.

Some questions I have for further development/improvements:
1. Should the tests cover every possible input? `LinuxDesktopEnvironmentResolver.test.ts` only tests for 4 different desktops as there are a dozen or two different strings to test for.
2. `LinuxApplicationRepository` actually works on every Linux desktop, it's just the icon generation that needs individual desktop support. Should the plugin support all desktops and default to the fallback icon for the few desktops that don't have icon generation support?

oliverschwendener:
1. I hoped that you would complete the rest of the desktop environments in `LinuxDesktopEnvironmentResolver.ts` and `LinuxDesktopEnvironmentResolver.test.ts`.
2. Yes, that's exactly what I would do.

src/main/Extensions/ApplicationSearch/ApplicationSearch.ts
```diff
@@ -36,7 +36,17 @@ export class ApplicationSearch implements Extension {
...
+            Linux: () => {
+                const supportedLinuxDesktopEnvironments: LinuxDesktopEnvironment[] = [
+                    "GNOME",
+                    "KDE",
+                    "Cinnamon",
+                    "MATE",
+                    "XFCE",
+                    "Pantheon",
+                ];
+                return supportedLinuxDesktopEnvironments.includes(this.linuxDesktopEnvironmentResolver.resolve());
```
oliverschwendener, line 48: If this works on every desktop environment, then we don't need this list here, right?

src/main/Extensions/ApplicationSearch/Linux/LinuxApplicationRepository.ts
```diff
@@ -60,7 +60,7 @@ export class LinuxApplicationRepository implements ApplicationRepository {
             (await this.fileSystemUtility.readFile(filePath)).toString(),
         )["Desktop Entry"];
         // https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys
-        const desktopEnv = this.environmentVariableProvider.get("XDG_CURRENT_DESKTOP")?.split(":");
+        const desktopEnv = this.environmentVariableProvider.get("ORIGINAL_XDG_CURRENT_DESKTOP")?.split(":");
```
oliverschwendener, line 63: Can we not just use the `LinuxDesktopEnvironmentResolver` here?

src/main/Core/ImageGenerator/Linux/LinuxAppIconExtractor.ts
```diff
@@ -273,15 +275,29 @@ export class LinuxAppIconExtractor implements FileIconExtractor {
...
+            Pantheon: () =>
+                this.commandlineUtility.executeCommand("gsettings get org.gnome.desktop.interface icon-theme"),
         };
 
-        return (await iconThemeNameMap[currentDesktopEnvironment]()).trim().slice(1, -1);
+        if (!iconThemeNameMap[this.linuxDesktopEnvironmentResolver.resolve()]) {
+            throw new Error(
+                `No available function to read icon theme for this desktop environment. Desktop: ${this.linuxDesktopEnvironmentResolver.resolve()}`,
+            );
+        }
```
oliverschwendener, line 294: In my opinion we should not even instantiate this class if the current desktop environment is not one of those defined here. Then we don't need this error handling here. What do you think?

src/renderer/Core/getImageUrl.ts
```diff
@@ -1,7 +1,5 @@
 import type { Image } from "@common/Core/Image";
 
 export const getImageUrl = ({ image, shouldPreferDarkColors }: { image: Image; shouldPreferDarkColors: boolean }) => {
-    return shouldPreferDarkColors
-        ? (image.urlOnDarkBackground ?? image.url)
-        : (image.urlOnLightBackground ?? image.url);
+    return shouldPreferDarkColors ? image.urlOnDarkBackground ?? image.url : image.urlOnLightBackground ?? image.url;
```
oliverschwendener: ?

src/main/Extensions/ApplicationSearch/Linux/LinuxApplicationRepository.ts
ke1v: Some Linux desktop contain other names besides just their desktop (ie. Ubuntu using "ubuntu:GNOME") and even though the desktop environment is just GNOME, we should still check "ubuntu" against the desktop file to determine whether or not we show it

oliverschwendener (changes requested):
See my comments

src/renderer/Core/getImageUrl.ts
oliverschwendener: Probably because I updated the prettier dependency in the meanwhile in `main`.

src/main/Core/ImageGenerator/Linux/SupportedDesktopEnvironments.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * A list of Linux desktop environments supported by ImageGenerator.
+ */
+export const SupportedLinuxDesktopEnvironment = ["Cinnamon", "GNOME", "KDE", "MATE", "XFCE", "Pantheon"];
```
oliverschwendener, line 4: Please use uppercase names only for classes, interfaces or types. And why not using the "LinuxDesktopEnvironment" type here?

```suggestion
export const supportedLinuxDesktopEnvironment: LinuxDesktopEnvironment[] = ["Cinnamon", "GNOME", "KDE", "MATE", "XFCE", "Pantheon"];
```

ke1v: I made the array to keep a list of supported desktops in one place (since we need it in `ImageGeneratorModule.ts` and `LinuxAppIconExtractor.ts`) to ensure better type safety. I updated it in the new commit but using Partial for `iconThemeNameMap` doesn't give as good type safety.

oliverschwendener: Having the array of supported linux desktop environments in one place makes sense here, I didn't see any problem with that.

oliverschwendener (approved):

docs/Extensions/ApplicationSearch/README.md
```diff
@@ -28,4 +28,10 @@ Supported operating systems:
...
 -   Windows
 -   macOS
--   Linux (with Cinnamon or Gnome Desktop environment)
+-   Linux (with the following desktop environments)
+    -   GNOME
+    -   KDE
+    -   MATE
+    -   XFCE
+    -   Cinnamon
+    -   Pantheon
```
oliverschwendener, line 37: Is this still true? Doesn't it work on all desktop environments now?

ke1v: App icons are only able to be generated on those desktops. Others will have the generic app icon `assets/Extensions/ApplicationSearch/linux-generic-app-icon.png` for every app. It's up to you to decide if that is considered fully functional

oliverschwendener: I would say the extension is fully functional when you can search and start apps. I added a note in the readme though that generic app icons will be used if the app icon can't be extracted.

oliverschwendener (approved):

PR #1160, opened 2024-08-13
Add a "simple folder search" extension

sashaweiss:
Adds a "Simple Folder Search" extension, similar to the functionality in `ueli` V8. Takes a list of paths from a persisted Settings extension (defaults to `["$HOME"]`, although screenshots include `["$HOME", "/"]` for variety), then searches them for directories and creates a `SearchResultItem` for each.

I use this feature non-stop in `ueli` V8, and since it was simple enough functionality I wanted to try reproducing it in V9. Please let me know what you think: is this is something you'd be interested taking? If so, please let me know if there are any stylistic or functional changes you'd like! (I tried to largely follow the conventions of existing extensions and settings, but I have very little experience with React so I'm fuzzy on its conventions.)

## Screenshots

image

image

## Style option: use abbreviated path for item name, vs. file name

I prefer this aesthetically, and it disambiguates between directories with the same name if they exist under multiple paths. However, because search result items are sorted alphabetically this results in all the simple-folder-search results being grouped together at the top of the list.

image


assets/Extensions/SimpleFolderSearch/macos-folder-icon.png
sashaweiss, line 1: Copied from the "Application Search" asset.

src/main/Extensions/SimpleFolderSearch/SimpleFolderSearch.ts
```diff
@@ -0,0 +1,121 @@
...
+            fileName: string;
+        };
+
+        const searchResultBatches: SearchResult[][] = await Promise.all(
+            this.getPathsToSearch().map(async (folderPath: string): Promise<SearchResult[]> => {
+                const directoryEntries = await this.fileSystemUtility.readDirectory(folderPath);
+                return directoryEntries.map((directoryEntry): SearchResult => {
+                    return {
+                        fullPath: directoryEntry,
+                        abbreviatedName: directoryEntry.replace(this.app.getPath("home"), "~"),
```
sashaweiss, line 60: I thought this looked nicer for display (even just in the description, i.e. "Open ~/Desktop") than the full path, although perhaps it's not conventional on Windows? Happy to adjust, for example to be platform-specific if you can suggest a better alternative for Windows.

src/main/Extensions/SimpleFolderSearch/SimpleFolderSearch.ts
```diff
@@ -0,0 +1,121 @@
...
+            (accumulator: SearchResult[], nextValue: SearchResult[]): SearchResult[] => {
+                return accumulator.concat(nextValue);
+            },
+            [],
+        );
+
+        // Filter out non-directories, and hidden items.
+        const filteredSearchResults: SearchResult[] = searchResults.filter((searchResult: SearchResult): boolean => {
+            return this.fileSystemUtility.isDirectory(searchResult.fullPath) && !searchResult.fileName.startsWith(".");
+        });
```
sashaweiss, line 77: Could make this configurable, although these are the settings I prefer and I figured for a V1 they were reasonable defaults.

src/renderer/Extensions/SimpleFolderSearch/SimpleFolderSearchSettings.tsx
sashaweiss, line 1: This was largely copied from the existing (and almost-identical) Application Settings.

src/main/Extensions/SimpleFolderSearch/SimpleFolderSearch.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    public getImage(): Image {
+        return this.folderIconImage;
+    }
+
+    public getI18nResources(): Resources<Translations> {
+        return {
+            "en-US": {
+                extensionName: "Simple Folder Search",
+            },
```
sashaweiss, line 118: I don't speak German, but if you can provide translations I'd be happy to add them here!

src/main/Extensions/SimpleFolderSearch/SimpleFolderSearch.ts
```diff
@@ -0,0 +1,121 @@
...
+        const filteredSearchResults: SearchResult[] = searchResults.filter((searchResult: SearchResult): boolean => {
+            return this.fileSystemUtility.isDirectory(searchResult.fullPath) && !searchResult.fileName.startsWith(".");
+        });
+
+        return filteredSearchResults.map((searchResult: SearchResult): SearchResultItem => {
+            return {
+                id: `${this.id}-path-${searchResult.fullPath}`, // Ensure unique
+                name: searchResult.fileName,
+                description: `Open ${searchResult.abbreviatedName}`,
+                image: this.folderIconImage,
```
oliverschwendener, line 84: Instead of having this static image you can also use the `FileImageGenerator` from `@Core/ImageGenerator` to get more dynamic icons for folders and files.

Just add the FileImageGenerator to the constructor and use it like this:

``` typescript
const images = await this.fileImageGenerator.getImages(searchResults.map(({ fullPath }) => fullPath));

return filteredSearchResults.map((searchResult: SearchResult): SearchResultItem => {
    return {
        id: `${this.id}-path-${searchResult.fullPath}`, // Ensure unique
        name: searchResult.fileName,
        description: `Open ${searchResult.abbreviatedName}`,
        image: images[searchResult.fullPath] ?? this.folderIconImage, // <= use the extracted icon and fallback to the generic image
        defaultAction: SearchResultItemActionUtility.createOpenFileAction({
            filePath: searchResult.fullPath,
            description: `Open ${searchResult.abbreviatedName}`,
        }),
    };
});
```

sashaweiss: Happy to include this change if you'd prefer stylistically – I think there is some value too in using the same blanket icon for all results for the extension, to visually separate them from, for example, application search results.

sashaweiss:
(Now with signed commits!)

oliverschwendener:
Thanks for your efforts! I started working on a very similar extension that not only allows you to search for folders but also for files (or both). If you want you can take a look at it here: #1194

In short: you this extension you can define folders that you want to search for files/folders. Per folder you can define if you want to do a recursive scan or just the first level.

sashaweiss:
Sounds good – thanks for the look @oliverschwendener.

PR #1161, opened 2024-08-13
Adds a setting for "max search results, when the search term is empty"

sashaweiss:
Adds a setting used to limit the max displayed search results, only when the search term is empty.

I appreciated the minimal, Spotlight-adjacent form factor of `ueli` in V8 (although V9 looks great!), and I don't often use the items in the first alphabetical page of search results that are shown when I launch `ueli`. I initially set out to entirely disable the initial page of search results, but as there was already a numerical setting for "max search results", I simply mirrored that setting and will plan to configure mine to be set to `0`. It's a little awkward having the empty window for search results, but (1) I don't know enough React to mess with that :) and (2) it does reduce flicker, which is nice.

Please let me know if this is something you'd like to take, and if there are any stylistic or functional changes you'd like me to make.

## Screenshots

| `maxSearchResultsEmptySearchTerm: 5` | `maxSearchResultsEmptySearchTerm: 0` |
|--------|--------|
| image | image |

image


src/renderer/Core/I18n/getCoreResources.ts
```diff
@@ -235,6 +236,7 @@ export const getCoreResources = (): { namespace: string; resources: Resources<Tr
                     automaticRescan: "Automatisches Neuscannen",
                     fuzziness: "Fuzziness",
                     maxResultLength: "Maximale Anzahl Suchergebnisse",
+                    // TODO: maxResultLengthEmptySearchTerm: "",
```
sashaweiss, line 239: Unfortunately I don't speak German, but happy to add in here if you can help translate!

src/renderer/Core/Search/SearchViewController.ts
```diff
@@ -67,6 +67,10 @@ export const useSearchViewController = ({
 
     const { value: fuzziness } = useSetting({ key: "searchEngine.fuzziness", defaultValue: 0.5 });
     const { value: maxSearchResultItems } = useSetting({ key: "searchEngine.maxResultLength", defaultValue: 50 });
+    const { value: maxSearchResultItemsEmptySearchTerm } = useSetting({
+        key: "searchEngine.maxResultLengthEmptySearchTerm",
+        defaultValue: 50,
```
sashaweiss, line 72: The default of 50 keeps behavior unchanged for users who prefer the current behavior, to keep the blast radius of this change minimal.

src/renderer/Core/Settings/Pages/SearchEngine/MaxResultLengthEmptySearchTerm.tsx
```diff
@@ -0,0 +1,29 @@
...
+        defaultValue: 50,
+    });
+
+    return (
+        <Setting
+            label={t("maxResultLengthEmptySearchTerm", { ns })}
+            control={
+                <Input
+                    value={`${maxResultLengthEmptySearchTerm}`}
+                    min={0}
```
sashaweiss, line 21: In contrast with the regular max results setting, which has a min of 1, this has a min of 0.

src/renderer/Core/Settings/Pages/SearchEngine/SearchEngine.tsx
```diff
@@ -28,6 +29,7 @@ export const SearchEngine = () => {
                 <RescanInterval automaticRescanEnabled={automaticRescanEnabled} />
                 <Fuzziness />
                 <MaxResultLength />
+                <MaxResultLengthEmptySearchTerm />
```
sashaweiss, line 32: Added to Search Engine settings, just below the existing max search results setting.

oliverschwendener:
While I appreciate your efforts, I don't see a benefit in the proposed solution. If I understand you correctly, you would appreciate to have an option to collapse the window when no search term is given, for example like this:

image

And when you start typing, extend the window height as necessary, for example like this:

image

Is that correct?

oliverschwendener:
As I don't see a value in this proposal I'll close this issue. Thanks for your efforts though :)

sashaweiss:
Hey @oliverschwendener!

> If I understand you correctly, you would appreciate to have an option to collapse the window when no search term is given ...
> ... And when you start typing, extend the window height as necessary ...

That's correct – my goal was to approximate that behavior without getting into the design complications that'd result. Apologies for the radio silence (been a busy summer), although I am planning to circle back to this change, and the others I opened! When I make the time, I'll look into a more complete solution involving dynamically sizing the results window.

PR #1172, opened 2024-08-21
feat: added linux desktop environment resolver

oliverschwendener:
@ke1v I added here the basic structure for a `LinuxDesktopEnvironmentResolver`. What do you think about this? Here we have a small module that is responsible of resolving the current Linux desktop environment. This can be then used by other modules like the `LinuxAppIconExtractor` or the `LinuxApplicationRepository`. You can target your implementation against this branch.

### DIFF
--- a/src/common/Core/LinuxDesktopEnvironment.ts
+++ b/src/common/Core/LinuxDesktopEnvironment.ts
@@ -1,4 +1,0 @@
-/**
- * A list of Linux desktop environments.
- */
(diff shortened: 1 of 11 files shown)


ke1v:
Thanks for the help. I've just started school again so it'll be a few days before I can integrate this and finish #1159.

oliverschwendener:
I'll merge this now, so you can add more desktop environments in your PR.

PR #1176, opened 2024-08-26
fix: Use "utf8" encoding for Buffer on Base64 conversion

ykhrito:
Encoding specified for Buffer on Base64 Conversion should be “utf8”, not “binary”.
This causes that non-ASCII characters, Emoji, etc... are not encoded/decoded correctly.

For example: "thumbs-up"👍should be encoded as "8J+RjQ=="

```shell
# using coreutils on Linux shell
$ printf 👍 | base64
8J+RjQ==
```

but Base64 Conversion result:
image

This PR fix would make the following result:
image

### DIFF
--- a/src/main/Extensions/Base64Conversion/Base64Converter.test.ts
+++ b/src/main/Extensions/Base64Conversion/Base64Converter.test.ts
@@ -6,11 +6,19 @@ describe(Base64Converter, () => {
         it("should return the correct base64 encoded value of a string", () => {
             expect(Base64Converter.encode("Test:1886")).toBe("VGVzdDoxODg2");
         });
+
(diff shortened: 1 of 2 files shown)


oliverschwendener:
@ChristopherSteiner

oliverschwendener:
@ykhrito Please add a test case for this.

ykhrito:
I've added test cases, please check it out.

oliverschwendener (approved):

oliverschwendener:
@ChristopherSteiner If this is good for you, I'll merge it.

ChristopherSteiner:
Yes looks good! Go ahead please

oliverschwendener:
@ChristopherSteiner Thanks for accepting the invite. Next time you get assigned as reviewer automatically.

PR #1181, opened 2024-08-30
fix(Hotkey): Support multiple modifiers

### LINKED ISSUE #1156: [Bug] V9, Ctrl+Shift+F2 not possible hotkey
Bug similar to #1150 

Previously, control + shift + F2 was possible in V8, and with what is written in electron documentation, it should be possible to do it, is it possible to fix it ?

# Preview

image

ke1v:
Fixes #1156

### DIFF
--- a/src/common/Core/Hotkey.test.ts
+++ b/src/common/Core/Hotkey.test.ts
@@ -30,12 +30,27 @@ describe(isValidHotkey, () => {
                 "Alt+]",
                 "Alt+(",
                 "Alt+)",
+                "Ctrl+Shift+Space",
+                "Alt+Shift+F2",
+                "Ctrl+Alt+Shift+Super+F8",
             ],
         }));
 
     it("should return false for all invalid hotkeys", () =>
         testIsValidHotkey({
             expected: false,
-            hotkeys: ["Space+Alt", "", " ", "abc", "F25", "Windows+Space", "Alt Space"],
+            hotkeys: [
+                "Space+Alt",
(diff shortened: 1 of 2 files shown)


src/common/Core/Hotkey.test.ts
```diff
@@ -30,12 +30,27 @@ describe(isValidHotkey, () => {
                 "Alt+]",
                 "Alt+(",
                 "Alt+)",
+                "Ctrl+Shift+Space",
+                "Alt+Shift+F2",
+                "Ctrl+Alt+Shift+Super+F8",
```
oliverschwendener, line 35: Are you shure that `Ctrl+Alt+Shift+Super+F8` is working?

oliverschwendener: See this article: https://www.electronjs.org/docs/latest/api/accelerator

ke1v: I tested it and it worked for me. I don't see anything from Electron docs that limit the number of modifiers keys?

oliverschwendener: True, just tested it myself. Somehow I thought this doesn't work.

oliverschwendener (approved):

oliverschwendener:
Looks good, thank you for the efforts!

oliverschwendener:
By the way, as you're now a collaborator on this repository, you can also create branches on this repo directly ;)

PR #1158, opened 2024-08-11
Add navigation with C-n and C-p

sashaweiss:
Hi! Excited to start using Ueli v9. I was missing navigating using `Ctrl-n` and `Ctrl-p`, so I made a change to add event handlers for those keys to the existing list of event handlers in `Search` that already listen for arrow keys.

Please let me know what you think.

PR #1183, opened 2024-09-01
Add navigation with C-n and C-p

sashaweiss:
Reopening per https://github.com/oliverschwendener/ueli/pull/1158, with signed commits!

> Not sure why I couldn't re-open the previous PR, but once I rebased with the signed commit the option was no longer visible.


oliverschwendener:
Cool thanks!

PR #1199, opened 2024-09-12
Add XML parser

scomans:
Adds XML reader function to file system utility used by updated Jetbrains Toolbox extension (#1200).

### DIFF
--- a/src/main/Core/XmlParser/Contract/XmlParser.ts
+++ b/src/main/Core/XmlParser/Contract/XmlParser.ts
@@ -1,0 +1,13 @@
+/**
+ * TODO: add description
+ */
+export interface XmlParser {
+    /**
+     * TODO: add description
+     *
+     * @param xml TODO: add description
+     * @param options TODO: add description
+     * @return TODO: add description
+     */
+    parse<T>(xml: string, options?: { preserveOrder: boolean; ignoreAttributes: boolean }): T;
+}
--- a/src/main/Core/XmlParser/Contract/XmlParser.ts
+++ b/src/main/Core/XmlParser/Contract/XmlParser.ts
@@ -1,13 +1,14 @@
 /**
- * TODO: add description
+ * Offers a method to parse XML content.
  */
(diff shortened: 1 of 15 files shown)


oliverschwendener:
I know that we have "readJsonFile" and "writeJsonFile" in the file system utility, but as we're introducing a new npm package here I would love to have the XML parsing in a separate module, similar to the `IniFileParser`. So if you have to read an xml file and parse it you would have to use the FileSystemUtility first to read the text file and then parse the XML using the XmlFileParser. Can you move the XML parsing out of the FileSystemUtility into it's separate module? Let me know if you need help setting up the module.

oliverschwendener (changes requested):
See my comment above

scomans:
Will do that tomorrow morning

oliverschwendener:
I added the basic structure how I think it would suit best. I would be happy if you could have a look and add some description for the XmlParser.

scomans:
Looks good to me

src/main/Core/XmlParser/Contract/XmlParser.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Offers a method to parse XML content.
+ */
+export interface XmlParser {
+    /**
+     * Parses the given XML string and returns the result as an object.
+     * @param xml - The XML string to be parsed.
+     * @param [options] - Optional settings for parsing.
+     * @param options.preserveOrder - Whether to preserve the order of elements.
+     * @param options.ignoreAttributes - Whether to ignore attributes in the XML.
+     * @returns {T} - The parsed result.
```
oliverschwendener, line 11: Very nice, thanks! 💪

oliverschwendener (approved):

oliverschwendener:
Would you like to be the code owner of this module? Then all future changes can be reviewed by you

scomans:
Sure :)

oliverschwendener (approved):

PR #1202, opened 2024-09-12
test: NodeJsFileSystemUtility

oliverschwendener:
This PR adds tests for the `NodeJsFileSystemUtility` which allows us to make changes in the future with higher confidence that we don't break anything.

### DIFF
--- a/src/main/Core/FileSystemUtility/NodeJsFileSystemUtility.test.ts
+++ b/src/main/Core/FileSystemUtility/NodeJsFileSystemUtility.test.ts
@@ -1,0 +1,325 @@
+import { mkdir, rm } from "fs";
+import { join } from "path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { NodeJsFileSystemUtility } from "./NodeJsFileSystemUtility";
+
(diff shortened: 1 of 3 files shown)


src/main/Core/FileSystemUtility/NodeJsFileSystemUtility.test.ts
```diff
@@ -0,0 +1,325 @@
...
+const deleteFolder = (path: string) =>
+    new Promise<void>((resolve, reject) => {
+        rm(path, { recursive: true }, (error) => {
+            if (error) {
+                reject(error);
+            } else {
+                resolve();
+            }
+        });
+    });
```
scomans, line 26: Why don't you use the promise based version?

oliverschwendener: Hehe good point. I'll adjust it.

oliverschwendener: Done.

scomans (approved):

PR #1203, opened 2024-09-12
Add commandline utility logging

IrishBruse:
Add some logging to view errors that happen with it and view any output that the command might have in the debug view


oliverschwendener:
Why just log errors in the `CommandlineActionHandler`? If another action handler wants to log errors too, then we need to add the same thing again. Here I have a counterproposal: #1205

PR #1205, opened 2024-09-13
feat(ActionHandler): log error when invoking action failed

oliverschwendener:
Title says it all.


IrishBruse (approved):
Yeah this looks like better way of doing

PR #1206, opened 2024-09-15
Add Japanese language support

deflis:
I am a Japanese user.
I created a Japanese locale because I want to use Ueli in Japanese.


oliverschwendener:
Thanks for your contribution! I can't validate the correctness of those translations, as I don't speak Japanese, so I just trust you that they're correct.

oliverschwendener (approved):

oliverschwendener:
Some of your commits are not signed. Can you please open a new PR with signed commits? Thanks.

PR #1210, opened 2024-09-16
:sparkles: Add option for setting a default target currency

### LINKED ISSUE #1207: [Feature Request] Default currency for currency converter
In v8 we could set the default currency, let's say EUR.

So instead typing:

```
10 usd to eur
```

it was enough to just type:

```
10 usd
```

That was super helpful and fast to convert currencies. Please consider adding back this option, thanks! 🙌

peterphmikkelsen:
# Description

Hello (again)! I am back with the second round of default target currency! (I was the person who implemented this for the old version as well) 😅

Still love the app and still use it very frequently 🤘 So kudos! 🙏 

Fixes #1207 

## :sparkles: Default Target Currency
I added the option for setting a default target currency. This includes updating the settings panel and a bit of the code to handle the input in the case of just writing the shorthand version.

The settings part let's the user pick only from the already selected list of currencies.

> [!note]
> _Note that I tried adding the Swiss German translation for the setting, but since I do not actually speak this language very well a quick review from someone who does would be appreciated_ 🙏

## 📷  How It Looks
image

## 📝 Notes
Couple of things of note:
- The chosen currency list from the settings does not seem to be used in any way in the extension (or I am just too stupid to understand it)
- The validation of the input got slightly more messy. I think it is tolerable but feel free to add a comment 👍
- When setting a default and then removing that currency from the `Currencies` list, the default still stays. Due to some weird state management in React I was not able to get around this properly.
- No testing code has been written. If this is required/desired I would need a bit of time to familiarize myself with how to do this. Manual testing has been done and seems successful.

### DIFF
--- a/src/main/Extensions/CurrencyConversion/CurrencyConversion.test.ts
+++ b/src/main/Extensions/CurrencyConversion/CurrencyConversion.test.ts
@@ -1,4 +1,5 @@
 import type { AssetPathResolver } from "@Core/AssetPathResolver";
+import type { SettingsManager } from "@Core/SettingsManager";
 import { describe, expect, it, vi } from "vitest";
 import { CurrencyConversion } from "./CurrencyConversion";
 import type { Rates } from "./Rates";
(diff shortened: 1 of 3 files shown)


peterphmikkelsen (commented):
Hey @oliverschwendener! Sorry for the linting issues 🤦‍♂️ I have resolved them now 👍

oliverschwendener:
Hey Peter, thanks for your efforts! :)

I just added some tests around the CurrencyConversion extension, so we can be sure that your addition works as expected and doesn't break the existing functionality. Please update your branch.

oliverschwendener (approved):
Well done, thank you for your efforts! 💪

Issue #1119, opened 2024-06-21
[Feature Request] Decouple Settings menu from Search menu


weiw11:
**Is your feature request related to a problem? Please describe.**
Search menu shares UI with settings menu, meaning if we need to adjust the settings menu to view more content, the search menu becomes hilariously huge.

Another issue is the `Hide On` option also applies to Settings UI.

**Describe the solution you'd like**
Search menu should be decoupled from Settings menu.

**Additional context**
![2024-06-21_13-56-02_firefox](https://github.com/oliverschwendener/ueli/assets/62213594/4ee46fa4-fd5d-4578-92ae-9e70756d633e)

oliverschwendener:
With the next release the settings will be in a separate window, see the attached example screenshot. Thanks for the patience.

![image](https://github.com/user-attachments/assets/e6a79fd2-8a81-4a88-b6c5-c1ead54b5236)

oliverschwendener:
Et voilà: https://github.com/oliverschwendener/ueli/releases/tag/v9.14.0

PR #1212, opened 2024-09-17
Add settings window resize

IrishBruse:
Added support to have different sized windows when in the settings window vs normal also added a setting to configure the sizes of them by default.
Some css changes might be a good idea as rn there is a slight amount of lag to the ui changing as it resizes maybe just set the whole screen to the background color temporarly and hide all the ui or do some fading not to sure.

image

image

This helps fix alot of the settings UI bugs where buttons and dropdowns are squashed or overflowing kinda of a patch to that bug but I plan on fixing that properly soon but this was a small addition that I extracted while messing around with that idea.


oliverschwendener:
Although I appreciate your efforts I think that this is not the best solution to the problem. I think what we really need is having the settings in a completely separate window. See #1119

IrishBruse:
All good was just messing around with some stuff 👍🏻

PR #1213, opened 2024-09-18
feature(cli-support): Toggle ueli using command line

### LINKED ISSUE #1170: [Bug] Global shortcut does not work on Linux using Wayland
**Describe the bug**
Global hotkeys are not available on Linux under GNOME Wayland sessions.

**To Reproduce**
1. Launch ueli on GNOME Wayland
2. Hide ueli window
3. Press hotkey
4. Ueli does not reappear

**Expected behavior**
Ueli should reappear when using the global hotkey

**Environment**
 - OS: [Linux]
 - Version 9.3.2

**Additional context**
This appears to be an upstream bug (https://github.com/electron/electron/issues/38288) in Electron, as Wayland does not provide a way for Electron to register global hotkeys.

ke1v:
Fixes #1170 by triggering the hotkeyPressed event when a second instance of ueli is ran


src/main/index.ts
```diff
@@ -1,13 +1,13 @@
...
-(async () => {
+const main = async () => {
     await Electron.app.whenReady();
 
-    Core.SingleInstanceLockModule.bootstrap(Electron.app);
+    const mitt = (await import("mitt")).default;
+    const platform = (await import("os")).platform;
+    const Core = await import("./Core");
+    const Extensions = await import("./Extensions");
+
```
ke1v, line 10: @oliverschwendener Is the switch to async imports here alright? Importing synchronously means ueli hangs for about 800ms (on my machine) before sending the second instance event to the first instance.

oliverschwendener: I did some performance testing using node's `perf_hooks` module. To me it does not look like the import takes so long.

- The `import` statements takes around 0.07ms
- Waiting for `app.whenReady()` takes around 40ms
- Waiting for `requestSingleInstanceLock` takes around 0.75ms
- Bootstrapping the electron modules takes around 0.75ms
- Bootstrapping the core modules takes around 40ms

oliverschwendener: Can you do some measuring and see if it behaves similarly on your machine?

src/main/Core/SingleInstanceLock/SingleInstanceLockModule.ts
```diff
@@ -1,10 +1,13 @@
...
-            console.log("Quitting application. Reason: another instance is already running");
-            app.quit();
-        }
+    public static bootstrap(dependencyRegistry: DependencyRegistry<Dependencies>) {
+        const eventEmitter = dependencyRegistry.get("EventEmitter");
+        const app = dependencyRegistry.get("App");
+
+        app.on("second-instance", () => {
+            eventEmitter.emitEvent("hotkeyPressed");
+        });
```
oliverschwendener, line 11: I get what you're trying to achieve here but this is a very bad solution to the problem.

As the name of the event "hotkeyPressed" is pretty clear, **it should only be emitted when the hotkey is actually pressed**. Because that's not the case here this event should not be emitted.

ke1v: Ah sorry don't know how I managed to overlook that. I'll clean up the code

src/main/index.ts
ke1v: I tried two different methods (apologies in advance for the messy code)
- [ES6 Import](https://github.com/ke1v/ueli/blob/99661c30cddefb5fc0fc0241507641bb21fd60a5/src/main/index.ts)
    - **SingleInstanceLock**
     Starts ~750ms (Import time)
     Duration ~110ms (Electron start time)
- [Dynamic Import](https://github.com/ke1v/ueli/blob/f6df85858a86f39a48115db4f4ffc50a276281fb/src/main/index.ts)
    - **Import**
     Starts ~180ms
     Duration ~700ms
    - **SingleInstanceLock**
     Starts ~900ms
     Duration 0.3ms

Both tests seems to suggest that `import` takes ~700ms. `app.whenReady()` and `requestSingleInstanceLock` takes about 100ms and 0.3ms which lines up with your testing. How did you time the import statements? 

Tested on a Ryzen 2600

oliverschwendener:
I'm closing this PR due to inactivity. Feel free to reopen if you decide to work on this again.

ke1v:
Sorry about inactivity. Changes have been made and I've added a warning when Wayland is detected as Electron falsely returns `true` when registering a shortcut

oliverschwendener:
Yes, go for it! We can move it anytime if we think it's not the right place 👍

src/renderer/Core/I18n/getCoreResources.ts
```diff
@@ -55,6 +55,8 @@ export const getCoreResources = (): { namespace: string; resources: Resources<Tr
                     searchHistoryEnabledDescription: "The search history lets you see your previous searches",
                     searchHistoryLimit: "Search History Limit",
                     searchHistoryLimitHint: "Limits the maximum number of search history items",
+                    waylandWarning: "Wayland detected. Hotkeys may not work.",
```
oliverschwendener, line 58: Can you quickly explain what "wayland" exactly is? 😅 Why dont' we need the hotkey in this case?

src/main/Core/BrowserWindow/BrowserWindowModule.ts
```diff
@@ -178,6 +183,13 @@ export class BrowserWindowModule {
 
         ipcMain.on("escapePressed", () => shouldHideWindowOnEscapePressed() && browserWindowToggler.hide());
 
+        app.on(
+            "second-instance",
+            () =>
+                shouldToggleWindowOnSecondInstance() &&
```
oliverschwendener, line 189: Should this really be configurable? Why shouldn't we just always show/focus the already running instance when a second instance is launched?

src/main/Core/ExtensionManager/ExtensionManagerModule.ts
```diff
@@ -73,6 +74,10 @@ export class ExtensionManagerModule {
             },
         );
 
+        ipcMain.on("getEnvironmentVariable", (event, { environmentVariable }) => {
+            event.returnValue = environmentVariableProvider.get(environmentVariable);
+        });
```
oliverschwendener, line 79: Why did you place this in the `ExtensionManagerModule`? This has nothing to do with extensions. I suggest moving it to the appropriate module.

src/renderer/Core/I18n/getCoreResources.ts
ke1v: Wayland is a new window management system many Linux distros are switching to. Due to security, windows are no longer able to read keyboard inputs when not in focus, and thus, no way to listen for global hotkeys. There's a system API in the works to fix this (that Chromium merged last week), but it is still under development.

src/main/Core/BrowserWindow/BrowserWindowModule.ts
ke1v: I made it an option in case anyone who doesn't need this feature wants ueli to just silently quit the second instance the way it currently does. I can remove it if you'd like

src/main/Core/ExtensionManager/ExtensionManagerModule.ts
ke1v: Oops, forgot to move it after testing. Will fix rn

src/main/Core/BrowserWindow/BrowserWindowModule.ts
oliverschwendener: Yes, please remove it.

src/common/Core/ContextBridge.ts
```diff
@@ -28,6 +28,7 @@ export type ContextBridge = {
     getAvailableExtensions: () => ExtensionInfo[];
     getAvailableTerminals: () => Terminal[];
     getEnabledExtensions: () => ExtensionInfo[];
+    getEnvironmentVariable: (environmentVariable: string) => string;
```
oliverschwendener, line 31: This returns potentially `undefined`:

```suggestion
    getEnvironmentVariable: (environmentVariable: string) => string | undefined;
```

ke1v:
Should finally be good to go. Tested and working on Windows 10, Linux Mint, and Fedora. I don't have a macOS device to test on.

Also missing the de-CH translations as I don't speak German

oliverschwendener:
Thanks, I'll test it as soon as I have time and I'll add the german translations.

oliverschwendener:
Works as expected on macOS. I adjusted the UI to look a bit nicer, see this screenshot:

image

oliverschwendener (approved):

src/main/Core/BrowserWindow/BrowserWindowModule.ts
```diff
@@ -166,7 +166,7 @@ export class BrowserWindowModule {
 
         ipcMain.on("escapePressed", () => shouldHideWindowOnEscapePressed() && browserWindowToggler.hide());
 
-        app.on("second-instance", () => browserWindowToggler.toggle());
+        app.on("second-instance", () => browserWindowToggler.showAndFocus());
```
ke1v, line 169: @oliverschwendener Any reason for using `showAndFocus` instead of `toggle`? This change means that the workaround for Wayland is unable to close the window with the "hotkey"

oliverschwendener: Because I think it's generally better behavior if launching a second instance always shows the already running instance instead of toggeling. If you're using Wayland you can use the hotkey to launch a second instance and the  escape key to close the window. If that's not enough we can still pass an argument when launching the second instance, for example `--toggle` and then choose to toggle instead of showing the already running instance.

oliverschwendener: For example:

```typescript
app.on("second-instance", (_, argv) => {
    if (argv.includes("--toggle")) {
        browserWindowToggler.toggle();
    } else {
        browserWindowToggler.showAndFocus();
    }
});
```

ke1v: That makes sense. I've updated the wiki to reflect this

PR #1228, opened 2024-10-05
Add settings to show/hide the app icon in the dock on macOS

### LINKED ISSUE #1018: [Feature Request] On Mac hide from Dock
On Mac how about adding the option to hide the app in the Dock? 

By the way 'Show Tray Icon' needs to be renamed to 'Show Menu Bar Icon'.  The row of icons at the top on the right are Menu Bar Icons not the Tray.

The Dock is at the bottom.

benjamin-kraatz:
This PR adds a new settings entry to the Appearance settings. It toggles the visibility of the App Icon in the macOS dock. Defaults to false (hidden).

Fixes #1018


src/main/index.ts
```diff
@@ -68,6 +67,7 @@ import * as Extensions from "./Extensions";
     Core.DialogModule.bootstrap(dependencyRegistry);
     Core.TerminalModule.bootstrap(dependencyRegistry);
     Core.ExtensionRegistryModule.bootstrap(dependencyRegistry);
+    Core.DockModule.bootstrap(dependencyRegistry);
```
oliverschwendener, line 70: Why did you move it so far down? As far as I can see the Dock module just depends on `App`, `SettingsManager` and `EventSubscriber`. I suggest bootstrapping it right after `SettingsManager`, which is the required dependency that is bootstrapped last.

src/main/Core/Dock/DockModule.ts
```diff
@@ -1,7 +1,24 @@
...
+        const showAppIconInDock = settingsManager.getValue("appearance.showAppIconInDock", false);
+        this.toggleAppIconInDock(app, showAppIconInDock);
+
+        const eventSubscriber = dependencyRegistry.get("EventSubscriber");
+        eventSubscriber.subscribe("settingUpdated[appearance.showAppIconInDock]", ({ value }: { value: boolean }) => {
+            this.toggleAppIconInDock(app, value);
+        });
+    }
+
+    private static toggleAppIconInDock(app: Electron.App, value: boolean) {
```
oliverschwendener, line 17: Please use a variable name that is more descriptive.

```suggestion
    private static toggleAppIconInDock(app: Electron.App, showAppIconInDock: boolean) {
```

src/main/index.ts
benjamin-kraatz: Right, makes sense. Haven't thought about this. Thanks for the suggestion

src/renderer/Core/Settings/Pages/Appearance/Appearance.tsx
```diff
@@ -70,6 +71,7 @@ export const Appearance = () => {
         <SettingGroupList>
             <SearchBarSettings />
             <KeyboardShortcuts />
+            {contextBridge.getOperatingSystem() === "macOS" && <DockSettings />}
```
oliverschwendener, line 74: These settings should not be under "Appearance" as the appearance settings are all settings on how the UI inside of the app looks. The dock settings are more related to how the app integrates with macOS. Therefore I suggest putting these settings to "General".

src/renderer/Core/Settings/Pages/Appearance/DockSettings.tsx
```diff
@@ -0,0 +1,38 @@
+import { useContextBridge, useSetting } from "@Core/Hooks";
+import { Setting } from "@Core/Settings/Setting";
+import { SettingGroup } from "@Core/Settings/SettingGroup";
+import { Switch } from "@fluentui/react-components";
+import { useTranslation } from "react-i18next";
+
+export const DockSettings = () => {
+    const ns = "settingsAppearance";
+    const { t } = useTranslation();
```
oliverschwendener, line 9: You can initialize this already with the right namespace:

```suggestion
    const { t } = useTranslation("settingsAppearance");
```

Then you don't need to specify the namespace everytime when using the `t()` function.

src/main/Core/Dock/DockModule.ts
```diff
@@ -1,7 +1,24 @@
...
+            this.toggleAppIconInDock(app, value);
+        });
+    }
+
+    private static toggleAppIconInDock(app: Electron.App, showAppIconInDock: boolean) {
+        if (showAppIconInDock) {
+            app.dock?.show();
+        } else {
+            app.dock?.hide();
+        }
```
oliverschwendener, line 22: Instead of having a dedicated method for this (and having to pass the dependencies into it) I would move this into the bootstrap method to keep things as simple as possible:

```typescript
import type { Dependencies } from "@Core/Dependencies";
import type { DependencyRegistry } from "@Core/DependencyRegistry";

export class DockModule {
    public static bootstrap(dependencyRegistry: DependencyRegistry<Dependencies>) {
        const app = dependencyRegistry.get("App");
        const settingsManager = dependencyRegistry.get("SettingsManager");
        const eventSubscriber = dependencyRegistry.get("EventSubscriber");

        const setAppIconVisibility = () => {
            if (settingsManager.getValue("appearance.showAppIconInDock", false)) {
                app.dock?.show();
            } else {
                app.dock?.hide();
            }
        };

        setAppIconVisibility();

        eventSubscriber.subscribe("settingUpdated[appearance.showAppIconInDock]", () => setAppIconVisibility());
    }
}

```

src/main/Core/Dock/DockModule.ts
```diff
@@ -1,7 +1,22 @@
...
+        const settingsManager = dependencyRegistry.get("SettingsManager");
+        const eventSubscriber = dependencyRegistry.get("EventSubscriber");
+
+        const setAppIconVisibility = () => {
+            const showAppIconInDock = settingsManager.getValue("appearance.showAppIconInDock", false);
+            if (showAppIconInDock) {
+                app.dock?.show();
+            } else {
+                app.dock?.hide();
+            }
```
oliverschwendener, line 16: There is no reason to have this variable as it's just used once and `"appearance.showAppIconInDock"` is also pretty clear what value represents:

```suggestion
            if (settingsManager.getValue("appearance.showAppIconInDock", false)) {
                app.dock?.show();
            } else {
                app.dock?.hide();
            }
```

src/main/Core/Dock/DockModule.ts
```diff
@@ -1,7 +1,22 @@
...
+        };
+        setAppIconVisibility();
```
oliverschwendener, line 18: Please pay attention to details: add line breaks between a variable declaration and a function call:

```suggestion
        };

        setAppIconVisibility();
```

src/renderer/Core/Settings/Pages/General/General.tsx
```diff
@@ -13,6 +14,7 @@ export const General = () => {
                 <Language />
                 <HotKey />
                 <Autostart />
+                <DockSettings />
```
oliverschwendener, line 17: I liked the previous approach better. Please render this component only on macOS.

benjamin-kraatz: This requires the `useContextBridge` hook to be on the `General` component, which I wanted to avoid to keep the component clean. This is why I added a wrapper component that just renders nothing on systems other than macOS. I can change it, but thought it would make things a bit more organized

oliverschwendener: I don't think that having the contextBridge in `General.tsx` hurts. A component should generally not care about when it should be rendered or not. It's the parent's component responsibility to decide when it should render specific settings or not. There are already similar examples, take a look at `Window.tsx`.

benjamin-kraatz: Agree. Changed it

oliverschwendener (approved):
Looks good, thanks for your efforts!

oliverschwendener:
Unfortunately your commits are not signed. Can fix this?

oliverschwendener:
Sorry, it was not mentioned in the contribution docs, but I added it now.

benjamin-kraatz:
> Unfortunately your commits are not signed. Can you fix this?

Never did this before, shame on me. But I'll set it up quickly.

benjamin-kraatz:
Hope it worked?

Very sorry for all the trouble. This is one of my first PRs I've ever made.

oliverschwendener:
> Very sorry for all the trouble. This is one of my first PRs I've ever made.

No worries :) Everything worked now, thanks!

Oneechan69:
@benjamin-kraatz Where is the setting to hide Ueli from the dock?

PR #1233, opened 2024-10-21
feat: enabled typescript strict mode

oliverschwendener:
This PR enables `"strict": true` in `tsconfig.node.json`. Would be cool if every code owner could review their code, thanks!

This PR is still work in progress...

PR #1237, opened 2024-10-25
fix: Refactor for typescript strict mode

ke1v:
Fixes and improvements for changes made in #1233


src/main/Extensions/ApplicationSearch/Linux/LinuxApplicationRepository.ts
ke1v, line 1: Changes in this file are because not all `.desktop` files found are apps. Some are settings or OS tools, which shouldn't be launched by the user nor should ueli attempt to search for their icon

oliverschwendener (approved):
Looks good to me, although I think it's time to finally add some tests for these classes

oliverschwendener:
And if the class itself is doing too many things, then consider extracting some things into separate dedicated classes.

Issue #1188, opened 2024-09-10
[Feature Request] Restore the Ability to Add Custom Search Engines in Web Search


Ilya-Green:
**Is your feature request related to a problem? Please describe.**
In ueli version 9.4.0, I can't find an option to add a new search engine in the Web Search settings. In version 8.26.0, I could add any search engine I wanted, such as YouTube, GitHub, Wikipedia, Google Maps, etc. After updating to the latest version, this functionality seems to be missing.
 
**Describe the solution you'd like**
I would like to have the ability to add custom search engines in the new versions of ueli, just like it was possible in previous versions.

oliverschwendener:
The latest version includes a "Custom Web Search" extension, thanks to @NiewView

PR #1249, opened 2024-11-14
feat: added Custom Web Search extension

NiewView:
Implement custom web search in a similar way, it was implemented in the version v8.

example
settings

Implements #1188


oliverschwendener:
Nice, Thanks a lot! 💪   I'll have a look as soon as I have time.

src/main/Extensions/CustomWebSearch/CustomWebSearchExtension.ts
```diff
@@ -0,0 +1,135 @@
...
+    async getSearchResultItems(): Promise<SearchResultItem[]> {
+        // Custom search engines do not have static search results
+        return [];
+    }
+
+    public getInstantSearchResultItems(searchTerm: string): SearchResultItem[] {
+        const customSearchEngines = this.settingsManager.getValue<CustomSearchEngineSetting[]>(
+            `extension[${this.id}].customSearchEngines`,
+            this.getSettingDefaultValue("customSearchEngines"),
+        );
```
oliverschwendener, line 37: You can use `getExtensionSettingKey` from `@common/Core/Extension` to get the right setting key:
 
```suggestion
        const customSearchEngines = this.settingsManager.getValue<CustomSearchEngineSetting[]>(
            getExtensionSettingKey(this.id, "customSearchEngines"),
            this.getSettingDefaultValue("customSearchEngines"),
        );
```

src/main/Extensions/CustomWebSearch/CustomWebSearchExtension.ts
```diff
@@ -0,0 +1,135 @@
...
+
+        const searchInput = searchTerm.replace(selectedSearchEngine.prefix, "").trim();
+        const encodeSearchInput = selectedSearchEngine.encodeSearchTerm ? encodeURIComponent(searchInput) : searchInput;
+
+        return [
+            {
+                name: selectedSearchEngine.name,
+                description: `Search in ${selectedSearchEngine.name}`,
+                id: `${selectedSearchEngine.name}:instantResult`,
+                image: this.getImage(),
```
oliverschwendener, line 53: Instead of this static image we could use the favicon from the URL directly. Then the search results would look like this, what do you think?

image

You could do this by adding the `UrlImageGenerator` to the constructor:

```typescript
public constructor(
    private readonly operatingSystem: OperatingSystem,
    private readonly assetPathResolver: AssetPathResolver,
    private readonly settingsManager: SettingsManager,
    private readonly urlImageGenerator: UrlImageGenerator,
) {}
```

Then use it when creating the search result items:

```typescript
return [
    {
        name: selectedSearchEngine.name,
        description: `Search in ${selectedSearchEngine.name}`,
        id: `${selectedSearchEngine.name}:instantResult`,
        image: this.urlImageGenerator.getImage(new URL(selectedSearchEngine.url).origin),
        defaultAction: {
            handlerId: "Url",
            description: `Search in default Browser`,
            argument: selectedSearchEngine.url.replace("{{query}}", encodeSearchInput),
        },
    },
];
```

docs/Extensions/CustomWebSearch/README.md
```diff
@@ -0,0 +1,25 @@
...
+-   **Prefix**: The prefix, which should trigger the custom search engine
+-   **Url**: The URL, to which you want to be redirected. Use `{{query}}` placeholder, where the search term should be inserted
+-   **Encode search term**: Specifies, if the search term should be URL encoded (e.g. `%20` for a space)
+
+![Settings](settings.png)
+
+## About this extension
+
+Author: [NiewView](https://github.com/NiewView)
+Extension Icon: Designed by [OpenMoji](https://openmoji.org/) – the open-source emoji and icon project. License: [CC BY-SA 4.0](https://creativecommons.org/licenses/by-sa/4.0/#)
```
oliverschwendener, line 19: Please use bullet points here:

```suggestion
- Author: [NiewView](https://github.com/NiewView)
- Extension Icon: Designed by [OpenMoji](https://openmoji.org/) – the open-source emoji and icon project. - 
- License: [CC BY-SA 4.0](https://creativecommons.org/licenses/by-sa/4.0/#)
```

oliverschwendener (changes requested):
Looks very good overall, thanks for your efforts! I left some minor comments.

oliverschwendener:
It would also be nice to have some unit tests around the most important functionality of this extension.

docs/Extensions/CustomWebSearch/README.md
NiewView: I made a minor tweak to your suggestion, because the license only belongs to the Icon not the whole extension.

src/main/Extensions/CustomWebSearch/CustomWebSearchExtension.ts
NiewView: That's great, I thought it will need way more effort to get it. 
I used the resolved URL for image fetching, so it will also work, if the query is part of the domain name.

NiewView:
@oliverschwendener Thanks for the detailed review.
All points should be addressed by now

oliverschwendener (approved):
Nice, thank you!

oliverschwendener:
Unfortunately there are unsigned commits. Can you please update this PR with signed commits? Thanks.

NiewView:
Is there a way to sign already pushed commits or what would be the recommended way to handle this? I was not aware of this feature until now 🙈

PR #1265, opened 2024-11-29
refactor: removed unnecessary "continue" keyword

oliverschwendener:
This PR removes two unnecessary `continue;` statements.


ke1v:
@oliverschwendener Sorry for the late response but looks good to me

PR #1278, opened 2024-12-03
feat: settings window

oliverschwendener:
The goal of this PR is to solve #1119 and have a separate window for the settings.

image

Open tasks:
 - [x] Add ability to open settings from tray icon context menu
 - [x] Add ability to open settings from ueli extension search results
 - [x] Add ability to open settings from keyboard shortcut
 - [x] Add ability to open settings from "browse extensions"
 - [x] Prevent hiding of search window on blur when settings window is open and focused
 - [x] Add correct window icon
 - [x] Fix bug that prevents the theme from being reactive
 - [x] Fix bug that prevents labels to be updated in settings window when language is changed
 - [x] Reduce memory usage by making sure settings window is destroyed when not used
 - [x] Extract common window management behavior to separate module
 - [x] Test app in dev mode on:
   - [x] Windows
   - [x] macOS
   - [x] Linux
 - [x] Test packaged app on:
   - [x] Windows
   - [x] macOS
   - [x] Linux


ChristopherSteiner (commented):
The changes to the extension look good

oliverschwendener:
Sorry for not waiting for the review, but as this change touched every possible area I thought it doesn't make much sense to ask for a review from everyone. But thanks anyway!

PR #1279, opened 2024-12-03
feat(UUIDExtension): Added UUID Extension

ChristopherSteiner:
Added a new extension for creating UUIDs/GUIDs. Most of it is explained in the README.
Let me know if I have to change some things.


oliverschwendener:
Thanks for your contribution! I'll have a look at this as soon as I have time.

oliverschwendener:
I quickly checked it out and it works really nicely!

Here are a few thoughts:
- I would like to have labels/descriptions for these UI controls, so it's clear what they are referring to. Maybe "Version" and "Number of UUIDs" or something similar. What do you think?
  image
- This UI control does not update when I enter a number with my keyboard. Is that intended? It only works when pressing "arrow up" or "arrow down". See the attached screen recording.

  https://github.com/user-attachments/assets/12bd6038-8e66-412a-91e3-37c5378429f4

ChristopherSteiner:
Thank you for your input @oliverschwendener !
I'll add some labels for the two inputs, probablly with the same label text as in the settings.
Regarding your second point, I noticed that too and was thinking about creating a seperate issue for it. It seems to be a common problem of the SpinButton in Ueli. For example it behaves the same with the precision setting of the Calculator Extension. I was not able to find the cause of it, but maybe we should just change the SpinButtons to the control that is used for the Search Engine Rescan Intervall setting? 
image

By the way is it possible to display the InstantSearchResults above the "normal"? In my opinion it would be more intuitive to type guid and instantly be able to press enter for copying a single new one, now you have to navigate an item down first because the action on top would be to open the UI, which is probably the uncommon case. (Very nit-picky I know :D)

oliverschwendener:
> By the way is it possible to display the InstantSearchResults above the "normal"?

Yes, that would be possible, but I decided not to do that because for some extensions, especially the extensions that return an instant search result for every search term, like the web search, see this example:
image

In this case whatever I type, the search result `Search for <TERM>` is always on the top, which is not ideal.

I guess the right way would be to allow every extension to decide if the instant search result items go before or after the regular ones. I'll look into that as soon as I have time.

oliverschwendener:
> It seems to be a common problem of the SpinButton in Ueli

Ok then let's look at that issue separately.

ChristopherSteiner:
> > It seems to be a common problem of the SpinButton in Ueli
> 
> Ok then let's look at that issue separately.

I've created a new issue for this #1281

ChristopherSteiner:
I've added some labels, they might look a bit weird though when using the default width of ueli and switched to Input instead of SpinButtons, they behave better in my opinion (but not perfect, negative and decimal numbers can be entered). Let me know what you think about these changes.

oliverschwendener:
Regarding the placement of the instant search result items, see this PR: #1282

oliverschwendener:
I like the new labels, thanks for adding that. What about placing the label above the dropdown/input. In my opinion it looks better. What do you think?

Currently:
image

My suggestion:
image

oliverschwendener:
You can now update your branch with the latest `main` branch. Then you can specify if you want to place the search results before or after the regular results.

oliverschwendener:
Et voilà:

image

oliverschwendener (approved):
It works really nice now, thanks for your efforts!

PR #1282, opened 2024-12-04
refactor(InstantSearchResults): added the option to order instant search results

oliverschwendener:
The current problem is that all instant search results go after the oridnary search results. For most extensions I assume that is the right behavior but for others it would be convenient to place the instant search results at the beginning of the search result list. See [this conversation](https://github.com/oliverschwendener/ueli/pull/1279#issuecomment-2516601653) as an example.

The behavior of all existing extensions should be unchanged, as all instant search results go **after** the regular search results.

@IrishBruse @NiewView Please confirm that your extensions still behave as expected.

### DIFF
--- a/src/common/Core/ContextBridge.ts
+++ b/src/common/Core/ContextBridge.ts
@@ -1,6 +1,7 @@
 import type { IpcRenderer, OpenDialogOptions, OpenDialogReturnValue, OpenExternalOptions } from "electron";
 import type { AboutUeli } from "./AboutUeli";
 import type { ExtensionInfo } from "./ExtensionInfo";
+import type { InstantSearchResultItems } from "./InstantSearchResultItems";
 import type { OperatingSystem } from "./OperatingSystem";
 import type { SearchResultItem } from "./SearchResultItem";
 import type { SearchResultItemAction } from "./SearchResultItemAction";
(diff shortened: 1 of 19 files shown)


NiewView (dismissed):

IrishBruse:
Im seeing some duplication
image

This is the searchResult after typing `ray`
```
[
  {
    id: 'vscode-file:///home/econn/git/raylib',
    name: 'raylib',
    description: 'Folder',
    image: [Object],
    defaultAction: [Object]
  },
  {
    id: 'vscode-file:///home/econn/git/tray-notes',
    name: 'tray-notes',
    description: 'Folder',
    image: [Object],
    defaultAction: [Object]
  }
]
```

I suspect its due to this prior when typing `r`
My guess is its only removing the file one then it keeps duplicating itself?
```
    {
      id: 'vscode-file:///home/econn/git/core',
      name: 'core',
      description: 'Folder',
      image: [Object],
      defaultAction: [Object]
    },
    {
      id: 'vscode-file:///home/econn/git/core',
      name: 'core',
      description: 'File',
      image: [Object],
      defaultAction: [Object]
    },
```

IrishBruse:
Ah seems to be a bug in my extension as its creating duplicate ids but you may still want to handle duplicates in case ill send in a pr to fix the bug to my extension

IrishBruse (approved):

oliverschwendener:
> Ah seems to be a bug in my extension

@IrishBruse Does this mean it is also a problem in the current `main` branch? Or was the problem introduced with the change of this PR?

IrishBruse:
The bug is on main but it's different this makes it worse by duplicating the broken file\folder

PR #1283, opened 2024-12-04
[VSCode ext] Fix bug when filename and folder are the same

IrishBruse:
Fix bug when 2 entries where filename and folder are the same it will just open whatever the path actually is regardless of which one you press


oliverschwendener (approved):

oliverschwendener:
Thanks!

PR #1284, opened 2024-12-05
perf: improve launch performance

ke1v:
Import modules in entry point dynamically to reduce launch time
Allows #1213 to feel more like a hotkey

Currently in discussion...


oliverschwendener:
This looks good to me. I added some adjustments, let me know what you think. Let's test this in all platforms in dev mode and packaged.

oliverschwendener (approved):
I tested it on Windows, macOS and Ubuntu and it works as expected. Thank you for your contribution!

PR #1294, opened 2024-12-13
feat: update VSCode extension to handle macOS file URIs correctly

meghoshpritam:
This pull request includes updates to the `VSCodeExtension` class in the `src/main/Extensions/VSCode/VSCodeExtension.ts` file to handle command arguments more effectively based on the operating system. The changes ensure that file URIs are correctly formatted for macOS.

* Added logic to modify the `argument` variable to replace the file URI prefix when the operating system is macOS.
* Updated the `defaultAction` to use the newly defined `argument` variable instead of creating it inline.

### DIFF
--- a/src/main/Extensions/VSCode/VSCodeExtension.ts
+++ b/src/main/Extensions/VSCode/VSCodeExtension.ts
@@ -164,11 +164,6 @@ export class VSCodeExtension implements Extension {
             this.getSettingDefaultValue("command"),
         );
 
-        let argument = template.replace("%s", `${commandArg} ${uri}`);
-        if (this.operatingSystem === "macOS") {
-            argument = template.replace("%s", `${uri.replace(/^file:\/\//, "")}`);
-        }
-
         return {
             id: `vscode-${fileType}-${uri}`,
             name: recent.label ?? Path.basename(path),
@@ -177,7 +172,7 @@ export class VSCodeExtension implements Extension {
             defaultAction: {
                 handlerId: "Commandline",
                 description: `Open ${fileType} in VSCode`,
-                argument,
+                argument: template.replace(/"?%s"?/, `${commandArg} ${uri}`),
             },
         };
     }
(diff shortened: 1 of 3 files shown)


IrishBruse:
Do you have an example of what this fixes I use it on Mac every day without encountering any issues with the uri

meghoshpritam:
> Do you have an example of what this fixes I use it on Mac every day without encountering any issues with the uri

@IrishBruse 

I am encountering the following error while O open the vscode folder. I removed the "" in the final command.

[12/14/2024, 11:51:45 AM][ERROR] Error while invoking action: Warning: 'folder-uri file:///Users/computer/work/kw' is not in the list of known options, but still passed to Electron/Chromium.
image

oliverschwendener (changes requested):
Please revert the changes in `package.json` and `package-lock.json`

meghoshpritam:
> Please revert the changes in `package.json` and `package-lock.json`

@oliverschwendener the changes in `package.json` and `package-lock.json` are reverted.

src/main/Extensions/VSCode/VSCodeExtension.ts
```diff
@@ -172,7 +172,7 @@ export class VSCodeExtension implements Extension {
             defaultAction: {
                 handlerId: "Commandline",
                 description: `Open ${fileType} in VSCode`,
-                argument: template.replace("%s", `${commandArg} ${uri}`),
+                argument: template.replace(/"?%s"?/, `${commandArg} ${uri}`),
```
IrishBruse, line 175: This means optional " but thats already the case just remove the quotes from the template in the settings. Also send what template you are using

IrishBruse (changes requested):

src/main/Extensions/VSCode/VSCodeExtension.ts
IrishBruse: image
just remove the quotes from the template to achieve the same thing

IrishBruse:
: Warning: 'folder-uri file:///Users/computer/work/kw' is not in the list of known options, but still passed to Electron/Chromium. not sure how this would be happening as what it should be passing is '--folder-uri file:///Users/computer/work/kw'

PR #1308, opened 2025-01-07
Show indicator while rescanning

tkohlmeier:
This adds an indicator to the search window while rescanning. Ueli 8 had such an indicator and I still think it is usefule since rescanning takes some time.

grafik

### DIFF
--- a/src/main/Core/ExtensionManager/ExtensionManager.test.ts
+++ b/src/main/Core/ExtensionManager/ExtensionManager.test.ts
@@ -1,3 +1,4 @@
+import type { BrowserWindowNotifier } from "@Core/BrowserWindowNotifier/Contract/BrowserWindowNotifier";
 import type { Extension } from "@Core/Extension";
 import type { ExtensionRegistry } from "@Core/ExtensionRegistry";
 import type { Logger } from "@Core/Logger";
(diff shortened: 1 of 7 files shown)


oliverschwendener:
Thanks for your contribution! I agree that an indicator for rescans makes sense, but I don't think that we should do it (visually) like proposed here. I think the footer area is a bit overcrowded now, there are too many items with too little space between. Maybe we should remove the labels (show them only in a tooltip, together with the keyboard shortcut) and just have icons. Let's improve that in a separate branch though before we add the rescanning indicator.

Additionally, there are different rescan types:
- The periodic rescan that rescans all extensions
- The manual rescan for a single extension when clicking on "rescan" or when enabling an extension

Your implementation only covers the first rescan type and I think it would make sense to cover the second one as well, what do you think?

tkohlmeier:
Yes, I agree with both points. I'm going to create a new PR for moving the labels and keyboard shortcuts to a tooltip.
When that is done, how would you like the rescan indicator to look like? Just the spinner or with label?

tkohlmeier:
I rebased the branch on the tooltip branch, moved the notification to the ExtensionManager and also added it to the `populateSearchIndexByExtensionId` method. Now the indicator is also visible when a single extension is rescanned.

PR #1309, opened 2025-01-07
Add keyboard shortcut for rescanning

tkohlmeier:
This adds <kbd>F5</kbd> as a keyboard shortcut for rescanning. I think it is much more convenient than having to enter `rescan`.

### DIFF
--- a/src/main/Core/SearchWindow/SearchWindowModule.ts
+++ b/src/main/Core/SearchWindow/SearchWindowModule.ts
@@ -26,6 +26,7 @@ export class SearchWindowModule {
         const settingsManager = moduleRegistry.get("SettingsManager");
         const vibrancyProvider = moduleRegistry.get("BrowserWindowVibrancyProvider");
         const browserWindowRegistry = moduleRegistry.get("BrowserWindowRegistry");
+        const ueliCommandInvoker = moduleRegistry.get("UeliCommandInvoker");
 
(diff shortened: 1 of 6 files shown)


tkohlmeier:
Added an entry for the hotkey to the settings:

grafik

Do you think this should be editable?

Also improved some translations.

oliverschwendener:
Here are a few thoughts:
- I don't think that typing "rescan" is so unintuitive that it needs a keyboard shortcut. If you have to do a rescan very often, consider lowering the rescan interval or mark the "rescan extensions" search result as a favorite, then it shows up higher in the search results. Then, maybe you just have to enter "re" and press enter.
- "F5" is very Windowsy, please suggest appropriate keyboard shortcuts for the other platforms
- Be careful with the term "hotkey". I consider the "hotkey" to be the keyboard shortcut that can be pressed outside of the app. Keyboard shortcuts inside of the app I call "keyboard shortcuts". I suggest to rename the "hotkey" in this context to "keyboard shortcut".

oliverschwendener:
Ok also fair enough :)

What do you think about the following?
- Let's use `F5` for all platforms to not collide with electron's refresh keyboard shortcut
- Let's call it "Keyboard shortcut" in the settings so the term "hotkey" is used solely for the "global hotkey"

tkohlmeier:
Sounds good. Just one thing about Electron's refresh function: When we remove the menu in #1327, this shortcut won't be available anymore.

oliverschwendener (approved):

PR #1311, opened 2025-01-09
Improve responsiveness of app launches

IrishBruse:
When pressing enter on an item if the app is slow to launch it feels like nothing happened as ueli doesnt show any feedback this pr makes it so that ueli hides itself while launching the app if the app errors it will reappear


oliverschwendener:
Thanks for the PR, I appreciate the effort. I agree that this should be improved, but I disagree with the implementation. I think the direction of dependencies is bad. The action handlers should not know anything about browser windows. Every future action handler needs to implement this again, and this is a bad design in my opinion. Here is how I would do it: [#1312](https://github.com/oliverschwendener/ueli/pull/1312)

IrishBruse:
Totally agree my implementation felt a bit off while writing it that makes way more sense thanks

PR #1312, opened 2025-01-09
Alternative to "Improve responsiveness of app launches"

oliverschwendener:
This is an alternative implementation to #1311.

The goal of this PR is to optimistically hide the search window before the action invocation started. This leads to better responsiveness when invoking actions.

Instead of logging an error we can also show an error box. I think that would make more sense as we should make the user aware immidiately that something went wrong. This is an example of the error box on Windows: 
image


oliverschwendener:
@IrishBruse Would you mind reviewing this PR?

IrishBruse:
Linux apps work but testing it with the vscode extension it looks like ill need to add `hideWindowAfterInvocation: true,` ill create a quick pr for this

IrishBruse (approved):

PR #1314, opened 2025-01-11
Hide window after invocation of workflow extension

oscardianno:
Hello :)
I just updated to v9 and migrated my previous Shortcuts to Workflows, but I found out that the window remained open after launching them. Then, I found that a similar commit to fix the same behavior but in the VSCode Extension had recently been pushed.
Greetings!


oliverschwendener (changes requested):
Hi:) I agree that invoking the workflow should hide the search window, so thanks for fixing this. Please make sure that your commit is signed, otherwise I can't accept your PR.

oliverschwendener:
Closing due to inactivity. I added the proposed change here: #1318

PR #1315, opened 2025-01-12
Move search window footer button labels to tooltip

tkohlmeier:
As suggested in #1308, this PR moves the labels and keyboard shortcuts of the search window footer buttons to tooltips.

grafik

### DIFF
--- a/src/renderer/Core/Footer.tsx
+++ b/src/renderer/Core/Footer.tsx
@@ -11,7 +11,7 @@ export const Footer = ({ children, draggable }: FooterProps) => {
             className={draggable ? "draggable-area" : ""}
             style={{
                 flexShrink: 0,
-                padding: 10,
+                padding: 8,
                 gap: 10,
                 boxSizing: "border-box",
                 display: "flex",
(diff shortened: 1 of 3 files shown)


oliverschwendener:
I really appreciate your effort but I think I'll implement this myself as I have a few objections with this PR:
- I would like to have arrows on the tooltips and set static positioning attributes as the position of the element is always the same. This makes sure that they won't be cut off on the edges.
- I think we can remove the "showKeyboardShortcuts" setting value as it doesn't make sense to hide them anymore. The reason why I added it in the first place is to have a way to hide them after a user got used to using them and doesn't need a reminder anymore.
- I would like to have the "MoreVerticalFilled" icon for the actions menu as I think it looks better.
- I would like to keep the icon size at 14 as I think it looks better.
- I think we should keep the "default action" button with the "Enter" icon to give the user a hint what will happen when they press enter. It doesn't make sense to hide that info in a tooltip.
- I would like to have vertical dividers between the buttons on both sides for better separation.

Sorry for the hassle!

Here is a screenshow of how I would finally add the rescan indicator:
- "extra-tiny" spinner
- small label on the right

image

PR #1316, opened 2025-01-12
Only show icon buttons and hide keyboard shortcuts in tooltip

oliverschwendener:
Replacement of #1315 and prerequisite of #1308

image


tkohlmeier:
Looks better. Just one thing: I added this to align the keyboard shortcuts vertically: `marginTop: showKeyboardShortcuts ? 2 : 0`
I think without it it is quite noticeably not perfectly aligned.

oliverschwendener:
Thanks for the hint. The keyboard shortcuts are now vertically aligned 👍

tkohlmeier:
Just one more thing to consider: maybe the icons for "Open application" in the footer and action menu should be the same?

grafik

tkohlmeier (approved):

oliverschwendener:
> maybe the icons for "Open application" in the footer and action menu should be the same?

I'm not sure if these icons should be the same. The icon in the actions menu should indicate what the action does and the icon for the default action should indicate how it can be invoked. 🤔 But I also see that the inconsistency can be confusing. For now let's leave it as it is.

PR #1319, opened 2025-01-14
fix: map AltGr to Ctrl+Alt for Windows

### LINKED ISSUE #589: [Bug] AltGr + Space doesn't launch the app
Greetings! The app was working fine with the default keyboard shortcut (Alt + Space) but when i change it to AltGr + Space it doesn't launch. Already tested the other windows options and they did work

mmattes:
Windows maps AltGr to Ctrl+Alt why AltGr does not work. This fix replaces AltGr with Ctrl+Alt before regestering the hot key. 

Fixes #589

### DIFF
--- a/src/main/Core/GlobalShortcut/GlobalShortcutModule.ts
+++ b/src/main/Core/GlobalShortcut/GlobalShortcutModule.ts
@@ -12,7 +12,13 @@ export class GlobalShortcutModule {
         const hotkeyIsEnabled = () => settingsManager.getValue("general.hotkey.enabled", true);
 
         const registerHotkey = () => {
-            const hotkey = settingsManager.getValue("general.hotkey", "Alt+Space");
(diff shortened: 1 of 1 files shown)


oliverschwendener:
Thank you for your contribution! I don't think this is a good solution. I don't like showing `AltGr` to the user but registering `Ctrl+Alt` in the background. Why don't we just show a hint in the UI that explains the issue, something like this:

"If you want to use AltGr in the hotkey please use Ctrl+Alt instead."

What do you think?

oliverschwendener:
To me this looks like an upstream issue in electron.

mmattes:
I agree, are you still intrested in a hint or should AltGr as as  valid Key be removed?

oliverschwendener:
Good point, I think we should remove `AltGr` from the valid modifiers, at least on Windows, as it doesn't work anyway.

oliverschwendener:
Closing this in favor of #1325

PR #1320, opened 2025-01-15
feat: added option to choose color mode

### LINKED ISSUE #1115: [Feature Request] v9: way to set dark theme
Currently, there's no way to set dark theme on Ueli v9 if you're using default light theme on Windows.

Only way to get it right now is to enable Windows's dark theme.

OS: Windows 10.

mmattes:
Added a option to appereance to choose the color mode, default is system but users can explicitly choose dark or light mode. 

I am not a typescript expert and have my doubts if i used the settings propperly. Also there is quite likely room for improvement in my code. 

Fixes #1115


oliverschwendener (changes requested):
Before I look at the code, please make sure that all commits are signed.

mmattes:
The commit is signed now

oliverschwendener (changes requested):
Please make sure the code is formatted correctly with prettier, see the contribution guidelines: https://github.com/oliverschwendener/ueli/blob/main/CONTRIBUTING.md#coding-style

If you don't have prettier configured to auto-format the code in your editor, then you can use this command to do it:

```
$ npm run prettier:write
```

oliverschwendener:
Unfortunately this new option does not work well together with the background material (Windows) or vibrancy (macOS) option. Here is a screenshot from macOS where the system appearance is set to "light" and ueli's color mode is set to "dark":

image

oliverschwendener:
Here is my implementation that should solve the issue mentioned above: #1328

PR #1323, opened 2025-01-17
Add hotkey for show in file eplorer

tkohlmeier:
This adds the hotkey <kbd>CTRL</kbd>+<kbd>O</kbd> like in ueli 8 for show in file explorer. I use this a lot and the action menu is just too cumbersome for this.

### DIFF
--- a/src/renderer/Core/Search/Search.tsx
+++ b/src/renderer/Core/Search/Search.tsx
@@ -86,6 +86,17 @@ export const Search = ({
                 },
                 needsToInvokeListener: (keyboardEvent) => keyboardEvent.key === "Enter",
             },
+            {
+                listener: () => {
+                    const action = searchResultItems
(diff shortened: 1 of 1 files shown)


oliverschwendener:
Thanks for another contribution! :) What do you think about specifying the keyboard shortcut on the action itself? Then every action could have a keyboard shortcut and we wouldn't treat the "ShowItemInFileExplorer" in a special way.

oliverschwendener:
And it would be cool to see the keyboard shortcut in the actions menu, for example like this:

image

tkohlmeier:
That makes a lot of sense. So I'm going to close this in favor of #1324. I can continue working on that, if you like. I think I need to open a new PR in that case, right?

oliverschwendener:
> I think I need to open a new PR in that case, right?

Let's continue the work together on my branch. I think the concept works it just needs some refinement.

PR #1324, opened 2025-01-18
feat: add keyboard shortcuts for actions

oliverschwendener:
This is an alternative implementation for the problem mentioned in #1323.

|Windows|macOS|Linux|
|---|---|---|
|image|image|TBD|

This is currently just a POC and is work in progress. A lot of stuff needs to be tested/cleaned up.


tkohlmeier:
Cool, I can push into this branch.

Improvements:
- Refactored the KeyboardEventHandler to fix the type errors
- Pressing <kbd>Shift</kbd>+<kbd>Enter</kbd> invoked the `Open` and `Open as admin` actions. Fixed this by breaking the loop after the first to be invoked action.
- Improved layout by removing the padding from KeyboardShortcut and adding a style prop allowing for setting the padding only where necessary.
- Fixed unit tests

tkohlmeier:
I found one issue: <kbd>Ctrl</kbd>+<kbd>f</kbd> does not work for adding to or removing from favorites.

oliverschwendener:
> I found one issue: Ctrl+f does not work for adding to or removing from favorites.

Should be fixed now.

src/main/Extensions/ApplicationSearch/ApplicationSearch.ts
```diff
@@ -67,8 +67,8 @@ export class ApplicationSearch implements Extension {
             "en-US": {
                 extensionName: "Application Search",
                 searchResultItemDescription: "Application",
-                openApplication: "Open application",
-                openApplicationAsAdministrator: "Open application as administrator",
+                openApplication: "Open",
+                openApplicationAsAdministrator: "Open as admin",
```
oliverschwendener, line 71: @tkohlmeier FYI I shortened these because the action menu is getting a bit long with the additional keyboard shortcuts. Maybe we have to check more actions and shorten the translations.

src/common/Core/SearchResultItemActionUtility/createShowItemInFileExplorerAction.ts
```diff
@@ -13,4 +13,5 @@ export const createShowItemInFileExplorerAction = ({ filePath }: { filePath: str
     handlerId: "ShowItemInFileExplorer",
     fluentIcon: "DocumentFolderRegular",
     hideWindowAfterInvocation: true,
+    keyboardShortcut: "Ctrl+o",
```
oliverschwendener, line 16: @tkohlmeier Probably we have to pass the keyboard shortcuts from the outside as on macOS the keyboard shortcut will be `Cmd+o` or maybe something completely different.

src/renderer/Core/Search/Helpers/getSearchResultItemActionByKeyboardshortcut.ts
```diff
@@ -0,0 +1,45 @@
...
+    for (const action of actions) {
+        if (action.keyboardShortcut) {
+            if (keyboardEventMatchesKeyboardShortcut(keyboardEvent, action.keyboardShortcut)) {
+                return action;
+            }
+        }
+    }
+
+    return undefined;
+};
```
oliverschwendener, line 45: @tkohlmeier This needs proper testing as this is the centerpiece of this change.

oliverschwendener:
>Pressing Shift+Enter invoked the Open and Open as admin actions. Fixed this by breaking the loop after the first to be invoked action.

Should we remove the hard-coded keyboard event handler for pressing enter and add a keyboard shortcut on all default actions? That would streamline things quite a bit.

oliverschwendener:
@tkohlmeier I just pushed some changes, let me know what you think. Sorry if that potentially conflicts with something that you're working on locally.

ke1v (approved):

src/common/Core/SearchResultItemActionUtility/createShowItemInFileExplorerAction.ts
```diff
@@ -3,7 +3,13 @@ import type { SearchResultItemAction } from "@common/Core";
 /**
  * Creates an action to show the given file in the default file browser.
  */
-export const createShowItemInFileExplorerAction = ({ filePath }: { filePath: string }): SearchResultItemAction => ({
+export const createShowItemInFileExplorerAction = ({
+    filePath,
+    keyboardShortcut,
+}: {
+    filePath: string;
+    keyboardShortcut?: string;
```
tkohlmeier, line 11: Why is `keyboardShortcut` optional? Shouldn't we register a shortcut for the Simple File Search as well? It's the only occurrence without a shortcut. The same applies for other actions.

tkohlmeier (commented):
Improved some tests and added one comment. The rest looks good to me.

src/common/Core/SearchResultItemActionUtility/createShowItemInFileExplorerAction.ts
oliverschwendener: I decided to make it optional because if an extension wants to use this action as the default action it doesn't make sense to specify it here as it gets overwritten with "Enter" later on.

tkohlmeier (approved):

PR #1327, opened 2025-01-19
Remove settings window menu

tkohlmeier:
Currently in settings window, this menu can be opened by pressing <kbd>Alt</kbd>:

grafik

This PR prevents this by handling the <kbd>Alt</kbd> key up event.

### DIFF
--- a/src/main/Core/SearchWindow/SearchWindowModule.ts
+++ b/src/main/Core/SearchWindow/SearchWindowModule.ts
@@ -53,7 +53,7 @@ export class SearchWindowModule {
 
         searchWindow.setVisibleOnAllWorkspaces(settingsManager.getValue("window.visibleOnAllWorkspaces", false));
 
-        if (!import.meta.env.DEV) {
+        if (app.isPackaged) {
             searchWindow.removeMenu();
         }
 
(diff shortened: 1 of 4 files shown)


tkohlmeier:
You are right, that is much cleaner. One drawback: Opening the dev tools with <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>I</kbd> is no longer possible when the menu is removed (and all the other shortcuts, which is probably a good thing). During development, I sometimes use the dev tools. How should we handle this? Having the dev tools in the production version doesn't make sense but during development it should be available.

oliverschwendener:
Ok I didn't see that coming. I also heavly use the dev tools while developting. Maybe we can just remove the menus when the app is packaged.

tkohlmeier:
I am now using [Vite environment properties](https://vite.dev/guide/env-and-mode#node-env-and-modes) to call `removeMenu()` only in non-dev mode. I didn't use the `PROD` property as I think those Vite properties are not available at all when running from a packaged build. I couldn't check this though because I'm just getting a white screen when opening ueli from release\win-unpacked.

What do you think about this solution? Is there some class/module where we could put the mode info instead of accessing `import.meta.env.DEV` directly from the search and settings window module?

tkohlmeier:
Ah, now I found `app.isPackaged`. Using that now.

oliverschwendener:
By the way what exactly is the problem with having the menu in the settings window?

tkohlmeier:
I just don't think that anybody would use it. And it contains some items that don't make any sense, like the Help menu (see above) or `Toggle Developer Tools` which doesn't even work in the packaged version. But if you like the menu maybe we can just remove some items from it.

oliverschwendener:
Ok fair enough :)

Issue #1171, opened 2024-08-19
[Bug] Calculator precision option not working as expected


Mr-Gamer0:
**Describe the bug**

when I make the calculation
5050-15, I get the result 5040 instead of the correct 5035. so 5 goes off too little. but when I do 5050+15, I get 5070 instead of the correct 5065. so here i get 5 extra

**To Reproduce**
make the calculation in Ueli

**Expected behavior**
the correct solution of the calculation

**Screenshots**
not required 

**Environment**
 - OS: Windows 11 pro
 - Version 22H2

**Additional context**
not required

oliverschwendener:
Weird, I can't reproduce the issue. What Ueli version are you running? Can you show us the Calculator settings?

<img width="617" alt="image" src="https://github.com/user-attachments/assets/a774e9ab-efe3-4f22-b255-8d1cccd13699">

<img width="613" alt="image" src="https://github.com/user-attachments/assets/20a19163-c12d-432e-9434-2ef071940883">

Mr-Gamer0:
that si strange.
Here are some pictures.

![image](https://github.com/user-attachments/assets/f316565b-d0ff-4f7d-b078-1bef2dc36645)
![image](https://github.com/user-attachments/assets/d2c84f6d-09b5-4308-ba61-06d907d15c5f)
![image](https://github.com/user-attachments/assets/43c05c55-77c0-4b8d-a9d0-bb773acd68d7)

oliverschwendener:
Increase the precision to 8 then the issue should be gone. I'll look into it as soon as possible.

oliverschwendener:
I tried to fix it but struggled to find a solution that worked. As a workaround I suggest increasing the precision so that the results are correct again.

oliverschwendener:
This should be fixed with the next release.

PR #1345, opened 2025-02-03
Fixed bug with wrong results and low precision

MarcoSennHaag:
This could be a potential fix for bug #1171
Always calculate with high precision and format the result to the desired precision if it's numeric.


oliverschwendener:
Thanks for your contribution. I see that this fixes the reported issue, thanks for that! I completely misinterpreted mathjs' "precision" option.

The code is getting a bit out of thand though. I know the code was already ugly before but now it's getting even worse. For exmaple to me it's almost impossible to understand what's happening here:

```typescript
return isNaN(result) ? calculationResult : result.toFixed(precision).replace(/(\.0*|(?<=(\..*))0*)$/, "");
```

And with this fix only numeric results will be rounded (not sure if this was already a problem before) and results with units are still not rounded. For example this (I have set the precision to 2):

image

Maybe there is a way to configure mathjs to round it's results?

tkohlmeier:
If precision means the maximum number of decimal places, we should use BigNumber's `toDecimalPlaces()`, like this:

    const math = create(all, { precision, number: "BigNumber" });
    const result: BigNumber = math.evaluate(this.normalizeExpression({ expression, decimalSeparator, argumentSeparator }));
    return result.toDecimalPlaces(precision).toString();

MarcoSennHaag:
> If precision means the maximum number of decimal places, we should use BigNumber's `toDecimalPlaces()`, like this:
> 
> ```
> const math = create(all, { precision, number: "BigNumber" });
> const result: BigNumber = math.evaluate(this.normalizeExpression({ expression, decimalSeparator, argumentSeparator }));
> return result.toDecimalPlaces(precision).toString();
> ```

Thank's for your input, but it looks like it's not that easy, as math.evaluate can return up to 8 different result types and some tend to throw exceptions (with or without units). I have something in mind to try. I'll keep you updated.

tkohlmeier:
It's not always BigNumber although we pass `number: "BigNumber"` to create?

MarcoSennHaag:
> It's not always BigNumber although we pass `number: "BigNumber"` to create?

`number: "BigNumber"` is the format used when strings are parsed. But it has nothing to do with the result of the calculation which may or may not contain units, can be of type Complex number or Decimal. Arrays are also supported.

MarcoSennHaag:
I just pushed a new version which uses a Regex to parse the number out of the result, round it and joins together everything.
Please review @oliverschwendener and/or @tkohlmeier

oliverschwendener:
@MarcoSennHaag I just pushed another test use case, which I think is not covered by your fix. But in my opinion the current approach is going completely in the wrong direction. Instead of navigating our way around math.js, is there no way to configure it to return a rounded result?

@tkohlmeier Yes, the user setting "Precision" means decimal places. I completely misinterpreted that term from math.js.

MarcoSennHaag:
> @MarcoSennHaag I just pushed another test use case, which I think is not covered by your fix. If the unit contains a number, for example `m^2` for square meters, it doesn't work. See all supported units of math.js here: https://mathjs.org/docs/datatypes/units.html#prefixes.
> 
> In my opinion the current approach is going completely in the wrong direction. Instead of navigating our way around math.js, is there no way to configure it to return a rounded result?
> 
> @tkohlmeier Yes, the user setting "Precision" means decimal places. I completely misinterpreted that term from math.js.

Thanks for this test case! I changed the regex to support this as well. I'm also not 100% happy with this solution as math.js should do this correctly. `math.round` allows using precision (https://mathjs.org/docs/reference/functions/round.html), but I couldn't get it to work correctly (when units are part of the result). You can give it a try or maybe use this workaround here for most common cases (better than wrong results as before). You decide!

oliverschwendener (approved):
I'm going to approve this so we get correct calculation results again. But I still think it's terrible that we have to do this ugly workaround. I think if we just didn't figure out how to configure math.js properly.

PR #1347, opened 2025-02-05
Added instant base 64 conversion in search 

MarcoSennHaag:
Added instant base 64 conversion in search results for fast conversion (see readme)
Please review @oliverschwendener and @ChristopherSteiner


src/main/Extensions/Base64Conversion/Base64Conversion.ts
```diff
@@ -20,6 +26,44 @@ export class Base64Conversion implements Extension {
         private readonly translator: Translator,
     ) {}
 
+    public getInstantSearchResultItems(searchTerm: string): InstantSearchResultItems {
+        const results: Array<InvocationResult> = [];
+        if (searchTerm.toLowerCase().startsWith("b64e ") && searchTerm.length > 5) {
```
oliverschwendener, line 31: I'll let @ChristopherSteiner decide, but just wondering: why hardcoding the prefix and not making it configurable by the user?

src/main/Extensions/Base64Conversion/Base64Conversion.ts
```diff
@@ -20,6 +26,44 @@ export class Base64Conversion implements Extension {
         private readonly translator: Translator,
     ) {}
 
+    public getInstantSearchResultItems(searchTerm: string): InstantSearchResultItems {
+        const results: Array<InvocationResult> = [];
```
oliverschwendener, line 30: Just being nitpicky here: please try to stay consistent with the rest of the code base. Use `T[]` over `Array<T>` when possible:

```suggestion
        const results: InvocationResult[] = [];
```

src/common/Extensions/Base64Conversion/InvocationResult.ts
```diff
@@ -0,0 +1,4 @@
+export type InvocationResult = {
+    action: "encoded" | "decoded";
+    value: string;
+};
```
oliverschwendener, line 4: Please move this file to `src/main/Extensions/Base64Conversion`.

The `common` folder should only contain code that is shared between the renderer and the main process which is not the case here.

MarcoSennHaag: Ok, will move it

src/main/Extensions/Base64Conversion/Base64Conversion.ts
MarcoSennHaag: I like this idea for a configurable prefix. I will change it.

MarcoSennHaag: Implemented and good to review

ChristopherSteiner (dismissed):

src/main/Extensions/Base64Conversion/Base64Conversion.ts
```diff
@@ -66,6 +142,11 @@ export class Base64Conversion implements Extension {
                 copyToClipboard: "Copy result to clipboard",
                 encodePlaceHolder: "Enter your string to encode here",
                 decodePlaceHolder: "Enter your string to decode here",
+                encoded: "Encoded",
+                decoded: "Decoded",
+                encodeDecodePrefix: "Prefix used for encode and decode",
+                encodePrefix: "Prefix used for encode",
+                decodePrefix: "Prefix used for decode",
```
MarcoSennHaag, line 149:
```suggestion
                encodeDecodePrefix: "Prefix used for encoding and decoding",
                encodePrefix: "Prefix used for encoding",
                decodePrefix: "Prefix used for decoding",
```

src/common/Extensions/Base64Conversion/Base64ConversionSetting.ts
```diff
@@ -0,0 +1,5 @@
+export type Base64ConversionSetting = {
+    encodeDecodePrefix: string;
+    encodePrefix: string;
+    decodePrefix: string;
+};
```
oliverschwendener, line 5: Sorry, another detail here: I suggest renaming this to `Base64ConversionSettings` as this type represents as set of settings. I consider a "setting" (singular) to be a specific key-value pair, for example "encodeDecodePrefix" is the key, and "b64" is the value. A set of these key-value pairs, or in this case just the keys, is then called "settings" (plural).

src/main/Extensions/Base64Conversion/Base64Conversion.ts
```diff
@@ -1,15 +1,30 @@
...
+    createCopyToClipboardAction,
+    createEmptyInstantSearchResult,
+    createInvokeExtensionAction,
+    type InstantSearchResultItems,
+    type SearchResultItem,
+} from "@common/Core";
+import { getExtensionSettingKey } from "@common/Core/Extension";
 import type { Image } from "@common/Core/Image";
-import type { InvocationArgument } from "@common/Extensions/Base64Conversion";
+import type { InvocationArgument, Base64ConversionSetting as Settings } from "@common/Extensions/Base64Conversion";
```
oliverschwendener, line 14: Following up on my comment above: this is another inticator that this should be called `Base64ConversionSettings` as we rename it to `Settings` (plural) here.

src/common/Extensions/Base64Conversion/Base64ConversionSetting.ts
```diff
@@ -0,0 +1,5 @@
+export type Base64ConversionSetting = {
```
MarcoSennHaag, line 1:
```suggestion
export type Base64ConversionSettings = {
```

src/main/Extensions/Base64Conversion/Base64Conversion.ts
```diff
@@ -18,8 +33,69 @@ export class Base64Conversion implements Extension {
     public constructor(
         private readonly assetPathResolver: AssetPathResolver,
         private readonly translator: Translator,
+        private readonly settingsManager: SettingsManager,
     ) {}
 
+    public getInstantSearchResultItems(searchTerm: string): InstantSearchResultItems {
+        const encodeDecodePrefix: string = this.getSettingValue("encodeDecodePrefix");
+        const encodePrefix: string = this.getSettingValue("encodePrefix");
+        const decodePrefix: string = this.getSettingValue("decodePrefix");
```
oliverschwendener, line 42: I suggest doing this little typescript trick:

Adjust these two methods to the following:

```typescript
public getSettingDefaultValue<T extends keyof Settings>(key: T): Settings[T] {
    return this.defaultSettings[key];
}

private getSettingValue<T extends keyof Settings>(key: T): Settings[T] {
    return this.settingsManager.getValue(getExtensionSettingKey(this.id, key), this.getSettingDefaultValue(key));
}
```

Typescript will now know what the type will be based on the type definition in `Base64ConversionSettings` so you can remove the manual type hints here:

```suggestion
        const encodeDecodePrefix = this.getSettingValue("encodeDecodePrefix");
        const encodePrefix = this.getSettingValue("encodePrefix");
        const decodePrefix = this.getSettingValue("decodePrefix");
```

src/main/Extensions/Base64Conversion/Base64Conversion.ts
MarcoSennHaag, line 14:
```suggestion
import type { InvocationArgument, Base64ConversionSettings as Settings } from "@common/Extensions/Base64Conversion";
```

src/common/Extensions/Base64Conversion/Base64ConversionSetting.ts
MarcoSennHaag: Accepted

src/main/Extensions/Base64Conversion/Base64Conversion.ts
MarcoSennHaag: Accepted

Issue #1350, opened 2025-02-06
[Bug] CustomWebSearch doesn't trim entries


milnak:
As a follow up to my bug #1340 

Go to custom web search settings.

add a search engine.

give it a prefix that already exists (e.g. `"google"`) but add a trailing space `"google "`.

It will be permitted.

oliverschwendener:
@NiewView

oliverschwendener:
This should be fixed with v9.17.0: https://github.com/oliverschwendener/ueli/releases/tag/v9.17.0

Issue #1351, opened 2025-02-06
[Feature Request] CustomWebSearch settings - place "Add Search Engine" button above list


milnak:
perhaps at top right of window, e.g.

![Image](https://github.com/user-attachments/assets/cfbb8291-c4e2-432a-81b0-cffd73e45d7f)

The list grows over time requiring me to scroll all the way down to add a new entry.

it's a nit, but a feature request nonetheless.

oliverschwendener:
@NiewView

MarcoSennHaag:
@oliverschwendener: This issue can be closed, I guess

oliverschwendener:
Yes, thanks @MarcoSennHaag. This will be fixed with the next release.

MarcoSennHaag:
I think this was fixed in the 9.17.0 release already, @oliverschwendener

PR #1352, opened 2025-02-07
Fixed bug with untrimmed prefixes and moved Add button to top

MarcoSennHaag:
Fixed bug #1350 and implemented feature request #1351.

### DIFF
--- a/src/renderer/Extensions/CustomWebSearch/CustomWebSearchSettings.tsx
+++ b/src/renderer/Extensions/CustomWebSearch/CustomWebSearchSettings.tsx
@@ -72,6 +72,23 @@ export const CustomWebSearchSettings = () => {
     return (
         <SettingGroupList>
             <SettingGroup title={t("searchEngines")}>
+                <div style={{ marginTop: -32, marginBottom: 8, textAlign: "right" }}>
+                    <DialogTrigger disableButtonEnhancement>
+                        <Button onClick={() => openEditDialog()} icon={<AddRegular />}>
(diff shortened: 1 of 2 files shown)


NiewView (approved):

src/renderer/Extensions/CustomWebSearch/CustomWebSearchSettings.tsx
```diff
@@ -72,6 +72,23 @@ export const CustomWebSearchSettings = () => {
     return (
         <SettingGroupList>
             <SettingGroup title={t("searchEngines")}>
+                <div style={{ marginTop: -32, marginBottom: 8, textAlign: "right" }}>
```
oliverschwendener, line 75: The `marginTop: -32` seems like a very ugly workaround, please remove it. I guess you added it because of some unwanted spacing that is being applied by either `SettingGroupList` or `SettingGroup`? If that's the case then just don't use those components.

oliverschwendener (changes requested):
See my comment above.

src/renderer/Extensions/CustomWebSearch/CustomWebSearchSettings.tsx
MarcoSennHaag: I just submitted a hopefully nicer fix to get the same result without negative margins.

oliverschwendener (approved):

oliverschwendener:
@NiewView Can you please re-review this? Thanks!

PR #1355, opened 2025-02-08
Fixed invocation of Everything CLI

MarcoSennHaag:
Fixed invocation of Everything CLI (part of #503)

### DIFF
--- a/src/main/Extensions/FileSearch/Windows/EverythingFileSearcher.ts
+++ b/src/main/Extensions/FileSearch/Windows/EverythingFileSearcher.ts
@@ -23,8 +23,9 @@ export class EverythingFileSearcher implements FileSearcher {
             return [];
         }
 
+        const escapedSearchTerm = searchTerm.replace(/([\\&|><^])/g, "^$1");
         const stdout = await this.commandlineUtility.executeCommand(
-            `cmd /c chcp 65001>nul && "${everythingCliFilePath}" -max-results ${maxSearchResultCount} ${searchTerm}`,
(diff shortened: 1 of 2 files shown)


src/main/Extensions/FileSearch/Windows/EverythingFileSearcher.ts
```diff
@@ -23,8 +23,9 @@ export class EverythingFileSearcher implements FileSearcher {
             return [];
         }
 
+        const escapedSearchTerm = searchTerm.replace(/([\\&|><^])/g, "^$1");
         const stdout = await this.commandlineUtility.executeCommand(
-            `cmd /c chcp 65001>nul && "${everythingCliFilePath}" -max-results ${maxSearchResultCount} ${searchTerm}`,
+            `cmd /c chcp 65001>nul && "${everythingCliFilePath}" -max-results ${maxSearchResultCount} ${escapedSearchTerm}`,
```
oliverschwendener, line 28: Can you add a test for this so we can see how this should behave?

MarcoSennHaag: I need to try what's possible to do.

oliverschwendener: It would be nice to at least test the regex.

MarcoSennHaag: See latest commit

oliverschwendener (approved):
Very nice!

PR #1360, opened 2025-02-09
feat: add search result item customization

oliverschwendener:
This PR adds more customization options for the search result items:

- show icon (switch)
- show description (switch)
- description color (dropdown)
- name text size (slider)
- name text weight (dropdown)
- details text size (slider)
- details text weight (dropdown)


oliverschwendener:
Closed this as I think this is not really needed. Just wanted to try out something.

PR #1364, opened 2025-02-12
Add search filesystem to vscode extension

### LINKED ISSUE #1338: [Feature Request] Visual Studio Code extension search file path also
**Is your feature request related to a problem? Please describe.**

Currently vscode extension seems only search vscode recent history, but not actual file path.

eg. I have a folder `~/script`, and doesn't matter partial or full path, the search window will show nothing.

**Describe the solution you'd like**

vscode extension can search both vscode recent and actual path.

**Describe alternatives you've considered**

n/a

**Additional context**

Nothing show up with partial path:
Image

Still nothing with full path:
Image

IrishBruse:
This PR add the ability to open folder and files from Ueli in vscode other than just the recently opened stuff.

closes #1338

image


src/main/Extensions/VSCode/VSCodeExtension.ts
```diff
@@ -280,10 +290,50 @@ export class VSCodeExtension implements Extension {
...
+            before: [],
+        };
+    }
+
+    private getFilePathItem(path: string, template: string, isDir: boolean): SearchResultItem {
+        return {
+            id: `vscode-filepath-${path}`,
+            name: path,
+            description: isDir ? "Folder" : "File",
+            image: isDir ? this.getImage() : this.getDefaultFileImage(),
```
oliverschwendener, line 315: What do you think about showing only the filename/foldername in the `name` and adding the full path to the `details`?

Here is how that would look like with the "compact" search result list layout:
image

And here with the new "detailed" search result list layout:
image

IrishBruse: Yeah that makes way more sense thanks

src/main/Core/FileSystemUtility/NodeJsFileSystemUtility.ts
```diff
@@ -100,4 +100,9 @@ export class NodeJsFileSystemUtility implements FileSystemUtility {
         const fileNames = await readdir(folderPath, { recursive });
         return fileNames.map((fileName) => join(folderPath, fileName));
     }
+
+    public readDirectorySync(folderPath: string, recursive?: boolean): string[] {
+        const fileNames = readdirSync(folderPath, { recursive });
+        return fileNames.map((fileName) => join(folderPath, fileName.toString()));
+    }
```
oliverschwendener, line 107: Can you please add a test for this?

src/main/Extensions/VSCode/VSCodeExtension.ts
```diff
@@ -235,10 +237,18 @@ export class VSCodeExtension implements Extension {
...
 
-        const searchResultItems = this.recents.map((recent) => this.getSearchResultItem(recent, template, showPath));
-
         searchTerm = searchTerm.replace(this.getPrefix() + " ", "").trim();
 
+        if (!isPath(searchTerm)) {
+            return this.getRecentSearchResults(searchTerm, template, showPath);
+        } else {
+            return this.getFilesystemSearchResults(searchTerm, template);
+        }
```
oliverschwendener, line 246: Feel free to leave it as it is, I'm nit-picking here, but how about turning this into one single return statement? Don't you think that's much nicer to read?

```typescript
return isPath(searchTerm)
    ? this.getFilesystemSearchResults(searchTerm, template)
    : this.getRecentSearchResults(searchTerm, template, showPath);
```

IrishBruse: Personally I find brackets much nicer you end up just refactorying turnaries like that if you need to add something to one of the branches.

If its a simple value expression that doesnt line wrap I'll use them but other than that I usually stick with if else

oliverschwendener: Ok, valid point :)

oliverschwendener (approved):

PR #1365, opened 2025-02-13
Added switch for validation of 'invalid' uuids which are valid elsewhere

MarcoSennHaag:
Added switch for validation of 'invalid' uuids which are valid in other programming languages (e.g. 99edf227-d64c-5d90-d314-5d998c38c687)


oliverschwendener:
What do you mean with "valid in other programming languages"? I thought the UUID spec is not tied to any language? https://www.rfc-editor.org/rfc/rfc9562.html

MarcoSennHaag:
> What do you mean with "valid in other programming languages"? I thought the UUID spec is not tied to any language? https://www.rfc-editor.org/rfc/rfc9562.html



> 99edf227-d64c-5d90-d314-5d998c38c687

See this fiddle:
https://dotnetfiddle.net/0qXRA9

versus this one:
https://playcode.io/2259294

ChristopherSteiner:
This one might be related even if it doesn't contain a real solution
https://github.com/uuidjs/uuid/issues/742 

Seems GUID and UUID aren't entirely the same.

ChristopherSteiner (approved):

PR #1367, opened 2025-02-14
feat: drag & drop

oliverschwendener:
Work in progress...

### DIFF
--- a/src/common/Core/DragAndDrop.ts
+++ b/src/common/Core/DragAndDrop.ts
@@ -1,0 +1,3 @@
+export type DragAndDrop = {
+    filePath: string;
+};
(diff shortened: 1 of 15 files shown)


ke1v:
@oliverschwendener I just tested this on Linux Mint and I'm unable to drag/drop anything besides the icons of the search result. Have you tested this on another distro?

oliverschwendener:
Did you enable drag & drop in the settings? It's turned off by default.

ke1v:
Yes, I enabled it in the settings. There appears to be a dragging out animation, but trying to drop it anywhere doesn't do anything, while dragging out the icon of the search result will drop the file path of the icon in ueli's cache.

oliverschwendener:
I tested it yesterday and it doesn't work for me either. Maybe we have to disable drag & drop for linux applications again.

ke1v:
Made issue #1373 to track it. I'll take a look this weekend to see if I can fix it.

PR #1372, opened 2025-02-15
Fix deb icon

IrishBruse:
This fixes the deb .desktop file not sure why it wasnt working and app image still works fine
image

Tested by uninstalling and then reinstalling to verify


oliverschwendener:
I think it works now because of the size in the file name

*The icon filename must contain the size (e.g. 32x32.png) of the icon.*
https://www.electron.build/linux#icon.

Is there a reason why you want to change the app icon from "app-icon-dark.png"  to "app-icon-dark-transparent.png"?

IrishBruse:
Windows uses that icon right just was trying to to be consistent plus I don't really see icons with backgrounds much on linux

oliverschwendener:
It doesn't seem to work for me (using Ubuntu). Am I missing something?

image

IrishBruse:
Its the icon it adds in the "start menu" the deb file can't have an icon

electron-builder.config.js
```diff
@@ -48,7 +48,6 @@ const platformSpecificConfig = {
     linux: {
         ...baseConfig,
         linux: {
-            icon: "assets/Build/app-icon-dark.png",
```
oliverschwendener, line 51: Would it work if we add the path to the new PNG file?

IrishBruse: Pretty sure I tried that and it didn't work I'll give that another go after work

IrishBruse:
Before
image

After
image

PR #1383, opened 2025-03-01
fix: remove not working drag and drop for linux applications

oliverschwendener:
This PR removes the not working drag&drop for Linux applications. I think it doesn't make sense to have it until we know how to make it actually work.


ke1v (approved):
Agreed. I've tried looking into it but haven't been able to find the cause yet.

PR #1390, opened 2025-03-06
feat: allow custom settings file path

### LINKED ISSUE #1023: [Feature Request] sharing settings between computers, win/mac specially
The beauty of a cross-system app only makes sense if user can share settings between systems, so aliases, links, etc, are always updated.

Is there a folder settings that user can point somewhere else, like dropbox, google drive, nextcloud, and sync settings?

oliverschwendener:
Work in progress...

The goal of this PR is to have the option to set a custom settings file path.

Fixes #1023


MarcoSennHaag:
I see you write an additional config file for what I have tried to achieve. I like this idea (no merging required). How do you want to present this to the user?

One thing to consider (even not solved in my former pull request). What happens in this case:

PC 1: User changes some settings
PC 1: File gets uploaded to OneDrive (could be any service)
PC 2: Boots and starts Ueli (Autostart)
PC 2: Detects changes to config file in OneDrive and downloads new version
PC 2: How does Ueli react on this changed settings?

Should there be something like a FileSystemWatcher or how can we detect external settings/configuration changes?

oliverschwendener:
> PC 2: Detects changes to config file in OneDrive and downloads new version

PC 2 can't tell (and also doesn't care) if the settings file was changed by another system. When Ueli starts on PC 2 the settings will be read from the the updated settings file and it treats it the same way as if it was the standard settings file in the default location.

I think the only issue that would appear is if two Ueli instances are running at the same time on two different computers and they both modify the same config file (e.g. via OneDrive). In that case the last change to the settings file would be the one that takes priority. But I think this case can be neglected.

oliverschwendener:
> How do you want to present this to the user?

I think the only thing missing from this PR is that there should be an easy way to set the custom settings file path via the UI. Similar to the import dialog. I'll add this later as soon as I find time.

oliverschwendener:
I'm going to start over with this from scratch as I had a better idea how to implement this.

PR #1410, opened 2025-04-12
move config file to user data folder

crazy-max:
relates to https://github.com/oliverschwendener/ueli/pull/1407/files#diff-ca89e2b6a811d0bab309d87969163a1ba4d2f89911ef9110865a45e29dc8131bR14

Not sure if this is on purpose but I think `ueli9.config.json` should be in user data folder and not lost directly in home directory.

Also not sure what the purpose of this file is, it seems always empty:

```
$ cat ~/ueli9.config.json
{}
```

Settings are currently saved in `ueli9.settings.json` in my case.

### DIFF
--- a/src/main/Core/SettingsFile/Resolver/CustomSettingsFilePathResolver.test.ts
+++ b/src/main/Core/SettingsFile/Resolver/CustomSettingsFilePathResolver.test.ts
@@ -6,15 +6,15 @@ import { CustomSettingsFilePathResolver } from "./CustomSettingsFilePathResolver
 
 describe(CustomSettingsFilePathResolver, () => {
     const createCustomSettingsFilePathResolver = ({
-        homePath,
+        userDataPath,
         configFileExists,
         config,
     }: {
-        homePath: string;
+        userDataPath: string;
         configFileExists: boolean;
         config: Record<string, string>;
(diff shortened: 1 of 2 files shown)


oliverschwendener:
The config file and settings file serve different purposes:

- The settings file stores the user settings that can be modified in via the settings UI.
- The configuration file is optional and determines from which file the settings should be loaded. It is only used if you want to use a settings file that is stored somewhere differnt to the default location.

If you want to use a settings file different to the default location you can add the file path to the config file:

```json
{ "settingsFilePath": "<PATH_TO_YOUR_SETTINGS_JSON>" }
```

Now I see that the user data folder would have been better for the config file as well. The initial reason why I added it to the home folder is that it would be easy to set up manually, but because now you can do everything from the UI I don't see that that would be necessary. Unfortunately we can't (or shouldn't) change the behavior now as it would break the custom settings file paths for all users who already use the `ueli9.config.json` file in the home folder and I don't want to do that. Also if the file in the home folder is a problem for you just delete it, it's not used unless you want to have a custom file path to your settings.

crazy-max:
Indeed would be a breaking change, thanks for the explanation, that looks good enough to me.

PR #1412, opened 2025-04-14
Application Search: check folder exists on Windows

crazy-max:
Since v9 when a folder does not exist on Windows it fails to scan other folders while in v8 it was just a warning and skipped invalid paths. This might happen when for example you're using a USB drive that might not be plugged in. I think we should keep the same behavior. This is also consistent with behavior on Linux.

### DIFF
--- a/src/main/Extensions/ApplicationSearch/Windows/WindowsApplicationRepository.test.ts
+++ b/src/main/Extensions/ApplicationSearch/Windows/WindowsApplicationRepository.test.ts
@@ -1,5 +1,6 @@
 import type { Image } from "@common/Core/Image";
 import type { AssetPathResolver } from "@Core/AssetPathResolver";
+import type { FileSystemUtility } from "@Core/FileSystemUtility";
 import type { FileImageGenerator } from "@Core/ImageGenerator";
 import type { Logger } from "@Core/Logger";
 import type { PowershellUtility } from "@Core/PowershellUtility";
@@ -20,7 +21,7 @@ describe(WindowsApplicationRepository, () => {
             expected: WindowsApplication[];
             includeWindowsStoreApps: boolean;
         }) => {
-            const folders = ["folder1", "folder2"];
+            const folders = ["folder1", "folder2", "folder3"];
             const fileExtensions = ["ext1", "ext2"];
 
             const executeScriptMock = vi.fn().mockImplementation((s) => {
@@ -55,6 +56,12 @@ describe(WindowsApplicationRepository, () => {
 
             const fileImageGenerator = <FileImageGenerator>{ getImages: (f) => getImagesMock(f) };
 
+            const isDirectoryMock = vi.fn().mockImplementation((path) => ["folder1", "folder2"].includes(path));
+
+            const fileSystemUtility = <FileSystemUtility>{
+                isDirectory: (path) => isDirectoryMock(path),
+            };
+
             const warnMock = vi.fn();
 
             const logger = <Logger>{ warn: (m) => warnMock(m) };
@@ -68,6 +75,7 @@ describe(WindowsApplicationRepository, () => {
             const windowsApplicationRepository = new WindowsApplicationRepository(
                 powershellUtility,
                 settings,
+                fileSystemUtility,
                 fileImageGenerator,
                 logger,
                 assetPathResolver,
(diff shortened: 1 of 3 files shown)


src/main/Extensions/ApplicationSearch/Windows/WindowsApplicationRepository.ts
```diff
@@ -30,7 +32,14 @@ export class WindowsApplicationRepository implements ApplicationRepository {
...
     private async getManuallyInstalledApps(): Promise<WindowsApplication[]> {
-        const folderPaths = this.settings.getValue<string[]>("windowsFolders");
+        const folderPaths: string[] = [];
+        this.settings.getValue<string[]>("windowsFolders").forEach((folderPath) => {
+            if (this.fileSystemUtility.isDirectory(folderPath)) {
+                folderPaths.push(folderPath);
+            } else {
+                this.logger.warn(`${folderPath} does not exist or is not a folder.`);
+            }
+        });
```
oliverschwendener, line 42: Thanks for your contribution. Generally I'm happy with this change, but I would like to have a few adjustments:

1. Please extract the filtering of existing folder paths to a dedicated method
2. I would like to replace `.forEach()` with `for (const item of items)` for the following reasons:
  - Staying consistent with the rest of the codebase, we almost exclusively use `for .. of` loops (I know that the `LinuxApplicationRepository` also has a `.forEach` loop, I'll adjust this in a follow-up)
  - `for .. of `loops support `await` inside the loop
  - Easier to break, continue, or return
  - Slightly better performance in some cases
  - More familiar if coming from other programming languages
3. Please add a test that covers the issue that your change is fixing

crazy-max: Should be good

ke1v (approved):
Somehow I'm marked as a codeowner here despite not owning any of these files. 

Regardless, LGTM

src/main/Extensions/ApplicationSearch/Windows/WindowsApplicationRepository.test.ts
```diff
@@ -105,5 +114,44 @@ describe(WindowsApplicationRepository, () => {
                 ],
                 includeWindowsStoreApps: false,
             }));
+
+        it("should log a warning if a folder path does not exist or is not a directory", () => {
```
oliverschwendener, line 118: This test is insufficient, it just tests the behavior of the private method. I'd rather have a test that makes sure that the filtered folders get passed to the powershell script.

src/main/Extensions/ApplicationSearch/Windows/WindowsApplicationRepository.ts
```diff
@@ -17,11 +18,26 @@ export class WindowsApplicationRepository implements ApplicationRepository {
...
+        for (const candidate of candidates) {
+            if (this.fileSystemUtility.isDirectory(candidate)) {
+                existingFolderPaths.push(candidate);
+            } else {
+                this.logger.warn(
+                    `Unable to get applications from folder "${candidate}". Reason: path doesn't exist or isn't a folder`,
+                );
+            }
+        }
+        return existingFolderPaths;
```
oliverschwendener, line 38: We try to write code that is pleasant to read and easy on the eye. In my opinion this looks just too dense. Please add some space around the `for` loop:

```suggestion
        const existingFolderPaths: string[] = [];

        for (const candidate of candidates) {
            if (this.fileSystemUtility.isDirectory(candidate)) {
                existingFolderPaths.push(candidate);
            } else {
                this.logger.warn(
                    `Unable to get applications from folder "${candidate}". Reason: path doesn't exist or isn't a folder`,
                );
            }
        }

        return existingFolderPaths;
```

oliverschwendener (changes requested):
See my comments above

src/main/Extensions/ApplicationSearch/Windows/WindowsApplicationRepository.ts
crazy-max: Ok will check eslint/prettier to format that correctly.

src/main/Extensions/ApplicationSearch/Windows/WindowsApplicationRepository.test.ts
crazy-max: Sure

src/main/Extensions/ApplicationSearch/Windows/WindowsApplicationRepository.ts
crazy-max: Prettier does not seem to complain about current code style but I have adjusted based on your comment. You might need to adapt your rules so contributors are not confused in the future.

oliverschwendener (dismissed):

src/main/Extensions/ApplicationSearch/Windows/WindowsApplicationRepository.ts
oliverschwendener: I'll add a eslint rule in a follow-up

crazy-max:
Updated to fix ci issue:

image

oliverschwendener (approved):

PR #1425, opened 2025-05-01
Add support for extracting custom icon paths from .lnk files (resolves #1236)

### LINKED ISSUE #1236: [Bug] Icon Disappears or Displays Incorrectly When Target and Icon Files Differ in Shortcuts
**Describe the bug**
When the Target and Icon for a shortcut are not the same file, Ueli displays the icon from the Target instead of the one specified in the shortcut properties. This results in the icon either disappearing or showing incorrectly, as illustrated in the screenshots below.

**To Reproduce**
*No additional steps are needed if you already have shortcuts where the Target and Icon files differ.*
Otherwise, follow these steps:

- Create a shortcut where the Target and Icon files are different. (Refer to Fig 2 and Fig 4)
- In Ueli, go to *Settings* -> *Extensions* -> *Application Search*, and click *Rescan*.
- Check the search results in Ueli. You’ll see that the wrong icon is displayed or the icon is missing. (Refer to Fig 1 and Fig 3)

**Expected behavior**
Ueli should correctly display the icon that is specified in the shortcut's properties, even when it points to a different file than the Target.

**Screenshots**
*Fig 1*: The icon disappears or is incorrectly displayed when the Target and Icon files are different.

Fig 1

*Fig 2*: Properties of two files—the native Discord icon on the left displays correctly, while the custom icon I created (right) does not. I used `%LOCALAPPDATA%` to anonymize the path.

Fig 2

*Fig 3* shows search results in Ueli with an incorrect icon for the shortcut "LAV Video (MPC-HC internal)," where the Target is `mpc-hc64.exe` and the icon is `lavvideo.ax`.

Fig 3

Finally, *Fig 4* shows the properties of this shortcut, confirming that the icon file is not the same as the Target:

Fig 4

**Environment**
- OS: Windows 11 23H2 (22635.4371)
- Ueli Version: 9.8.0 (tested on Zip versions only)

**Additional context**
This issue is inherited from [#1133].

undecV:
#### Related Issue

- Issue: #1236

#### Summary of Changes
- Enhanced the `Get-Shortcut-Target` function to parse the `IconLocation` field of `.lnk` files.  
- If a valid icon path is found (e.g., `.ico`, `.dll`, or `.exe`), it will be used for icon extraction instead of the shortcut target.  
- If the icon path is not accessible, the function falls back to the shortcut target or the `.lnk` file itself as before.

#### Known Limitation

Although `IconLocation` strings may specify an icon index (e.g., `foo.dll,5`), PowerShell’s `[System.Drawing.Icon]::ExtractAssociatedIcon()` can **only extract index 0**.
Full support for non-zero icon indices would require calling native APIs like `ExtractIcon`, which is outside the scope of this script and PowerShell’s built-in capabilities.

#### Steps to Reproduce

1. Delete the cached icon folder:  `%AppData%\ueli\FileImageImageGenerator`
2. In Ueli, re-scan the **Application Search** extension.
3. Check that custom shortcut icons now render properly.

#### Screenshot (after fix)  

Screenshot


oliverschwendener:
Lovely, thanks a lot! I will test this as soon as I have time.

oliverschwendener:
In the meantime can check if the eslint warning is valid?

undecV:
This was a regex linting issue, which I overlooked at first. It’s now been corrected and double-checked to ensure it doesn’t affect functionality. My apologies for the oversight!

oliverschwendener (changes requested):
Thanks a lot for your efforts. Unfortunately this change causes other icons to have the ugly "shortcut" icon again :(

Before:
image


With this change:
image

Is there a way to fix this?

undecV:
I'm ashamed to say that I wasted your time—I’ve realized that I didn’t fully understand how this code and PowerShell work. Worse yet, my previous modifications had no actual effect and even broke the original behavior. 😞

I've now fixed the issue, and I’ve confirmed that in most cases, the extracted icon no longer includes the shortcut overlay.

In my tests, only folder icons still display the shortcut arrow, but this matches the previous behavior and may need to be addressed in the future.

Screenshot

oliverschwendener (approved):
Works like a charm, thanks!

Issue #1326, opened 2025-01-19
Settings window structure


tkohlmeier:
I have some suggestions/questions for the settings window structure:

- In the Window settings there is a section Appearance. There is also an Appearance section. I always confuse those, but I'm not sure how to structure it better. Maybe merge the two sections?
- Excluded items appear in the Search Engine settings. Favorites have there own section. Those settings appear very similar to me. Maybe we should move the favorites to Search Engine or the excluded items to a renamed Favorites section?
- Why is Hotkey in General settings and not in the Keyboard & Mouse section?

PR #1427, opened 2025-05-10
refactor: move excluded items to separate settings page

oliverschwendener:
This PR moves the excluded items to a separate settings page as suggested in #1326.

### DIFF
--- a/src/renderer/Core/Settings/Pages/index.tsx
+++ b/src/renderer/Core/Settings/Pages/index.tsx
@@ -1,6 +1,7 @@
 import {
     AppsAddIn20Regular,
     Bug20Regular,
+    EyeOffRegular,
     Info20Regular,
     Keyboard20Regular,
     PaintBrush20Regular,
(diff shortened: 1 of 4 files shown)


src/renderer/Core/Settings/Pages/index.tsx
```diff
@@ -71,6 +73,13 @@ export const settingsPages: SettingsPage[] = [
         element: <Favorites />,
         icon: <Star20Regular />,
     },
+    {
+        translation: { key: "title", namespace: "settingsExcludedItems" },
+        relativePath: "excluded-items",
+        absolutePath: "/excluded-items",
+        element: <ExcludedItems />,
+        icon: <EyeOffRegular />,
```
tkohlmeier, line 81: I think we should use `EyeOff20Regular` here.

tkohlmeier (changes requested):

src/renderer/Core/Settings/Pages/index.tsx
oliverschwendener: Good catch, thanks! I'll adjust it.

oliverschwendener: Done ✅

tkohlmeier (approved):
Looks good now, thanks!

Issue #1411, opened 2025-04-13
[Bug] Crash when calculating factorial of a decimal number


marginal23326:
**Describe the bug**
The ueli application encounters an uncaught exception and crashes when attempting to calculate the factorial (`!`) of a non-integer (decimal) number. Factorial calculations for integers work as expected.

**To Reproduce**
1.  Open ueli.
2.  Go to `Settings` > `Extensions`.
3.  Ensure the `Calculator` extension is enabled. If not, enable it.
4.  Return to the main ueli input field.
5.  Type a decimal number (e.g., `2.5`, `3.14`, `10.2`).
6.  Type the factorial symbol (`!`) immediately after the number (e.g., `2.5!`).
7.  Observe the application crash and display an error message window (see screenshot below).

**Expected behavior**
The application should not crash. Ideally, it should either:
*   Display a user-friendly error message within the ueli interface indicating that the factorial function is only defined for non-negative integers.
*   Or, simply ignore/skip the factorial calculation for non-integer inputs.

**Screenshots**
Here is the error message window that appears:

![Image](https://github.com/user-attachments/assets/ae254033-8d24-4ffe-a805-b2019f75d4ca)

**Environment**
 - OS: Windows 11 Pro
 - OS Version: 23H2 (OS Build 22631.5189)
 - Ueli Version: 9.21.0

oliverschwendener:
The next release will include a fix that will prevent the app from crashing, thanks to @mszu. "Factorial calculations for integers" will still not work though, we need to fix that separately.

oliverschwendener:
This should be fixed with the latest release, thanks to @MarcoSennHaag.

PR #1440, opened 2025-06-05
fix: recover from uncaught extension errors

mszu:
If any extension raises an uncaught exception while generating search results, the user is presented with an ugly stack trace and the main UI stops updating (see e.g. #1411):

error

This PR simply catches exceptions that surface while iterating over extensions during search result retrieval and logs the errors instead of dying. No results are displayed from the offending extension:

success_log

success


src/main/Core/ExtensionManager/ExtensionManager.test.ts
```diff
@@ -106,4 +106,48 @@ describe(ExtensionManager, () => {
...
+            getInstantSearchResultItems: (s): InstantSearchResultItems => {
+                getInstantSearchResultItemsMock(s);
+                throw new Error("simulated error");
+            },
+        };
+
+        const extensionRegistry = <ExtensionRegistry>{ getAll: () => [extension1, extension2] };
+        const getValueMock = vi.fn().mockReturnValue([extension1.id, extension2.id]);
+        const settingsManager = <SettingsManager>{ getValue: (key, defaultValue) => getValueMock(key, defaultValue) };
+        const mockedLogger = <Logger>({ error: vi.fn() } as unknown);
```
mszu, line 137: The `as unknown` intermediate cast is to avoid having to mock the unused debug/info/warn logger functions

oliverschwendener (approved):
Lovely, thanks alot!

PR #1446, opened 2025-07-02
feat: Hide initial search list

### LINKED ISSUE #1173: [Feature Request] Add option to diable/hide initial list of search results
**request**
Option to disable/hide list of search results that show up automatically when invoking Ueli. 

**solution**
Add toggle option in settings to disable this new feature that you have added.
So that when we invoke Ueli is just shows the search bar without any list until we search, just like the previous version of Ueli

Thanks you!!!

ke1v:
Adds the option to display only the search bar when empty. It currently hides or displays the entire search window, but I plan on making the height dynamic to the number of search results as it was in v8.

Fixes #1173

Depends on #1445


oliverschwendener:
See my comment here: https://github.com/oliverschwendener/ueli/pull/1445#issuecomment-3645887433

PR #1460, opened 2025-09-28
feat(browser-bookmarks): add real Chrome profile detection and selection in bookmarks extension

### LINKED ISSUE #1443: [Bug] Cannot import bookmarks when multiple Chrome profiles are present
**Describe the bug**  
When multiple Google Chrome profiles are present on a system, ueli is unable to import bookmarks correctly. It either fails to import any bookmarks or imports from the wrong Chrome profile.

**To Reproduce**  
1. Open Google Chrome.
2. Create multiple profiles (e.g., "Default", "Profile 1", "Profile 2").
3. Add different bookmarks to each profile.
4. Open ueli and enable the bookmarks plugin (if not already enabled).
5. Try to use ueli to search for bookmarks.

**Expected behavior**  
Ueli should import bookmarks from the correct or selected Chrome profile. Ideally, the user should be able to select which profile to import bookmarks from.

**Screenshots**  

Image

Image

**Environment**  
- OS: Windows 11  
- Version: 9.24.0  

**Additional context**  
Chrome stores bookmarks in separate folders for each profile under:  
`%LOCALAPPDATA%\Google\Chrome\User Data\`  
Each profile (e.g. "Default", "Profile 1") has its own `Bookmarks` file. It seems ueli only looks at the "Default" profile. Support for multi-profile detection or manual profile selection would resolve the issue.

TheCardios:
This PR adds support for detecting and selecting Google Chrome profiles in the Browser Bookmarks extension. 
- The settings UI now lists all available Chrome profiles found on the user's system (Windows/macOS).
- The user can select which Chrome profile to use for importing bookmarks.
- The backend and preload bridge have been extended to allow the renderer process to query system paths and read directories securely.
- The bookmarks import logic now uses the selected profile, ensuring correct and expected behavior for multi-profile Chrome setups.

Closes #1443

### DIFF
--- a/src/common/Core/ContextBridge.ts
+++ b/src/common/Core/ContextBridge.ts
@@ -45,6 +45,16 @@ export type ContextBridge = {
     getExtensionAssetFilePath: (extensionId: string, key: string) => string;
     getExtensionResources: <T extends Translations>() => { extensionId: string; resources: Resources<T> }[];
     getExtensionSettingDefaultValue: <Value>(extensionId: string, settingKey: string) => Value;
+
+    /**
+     * Restituisce il percorso di sistema richiesto ("home" o "appData").
+     */
+    getPath: (type: "home" | "appData") => Promise<string>;
+
+    /**
+     * Legge le sottocartelle di una directory.
+     */
+    readDir: (dirPath: string) => Promise<string[]>;
     getFavorites: () => string[];
     getInstantSearchResultItems: (searchTerm: string) => InstantSearchResultItems;
     getLogs: () => string[];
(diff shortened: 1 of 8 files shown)


oliverschwendener:
Thanks for your contribution. Unfortunately I can't accept this PR for different reasons:
- There are comments in a language different than English.
- In my opinon this could have been achieved (and should) by just adjusting the Browser Bookmarks extension-related files, without having to touch `ContextBridge`, `src/main/index.ts` and `src/preload/index.ts`. There is no need to expose more things to the renderer process. This can all be done on the main process within the Browser Bookmarks extension.
- Unnecessary imports in `src/main/index.ts` and incorrect placement of `ipcMain.handle("getPath", ...)` and `ipcMain.handle("readDir", ...)`. This should all be placed in some module.
- I would love to see the support for profiles for all chromium-based browsers, because (I assume) they all handle profiles in a similar or even the same way. If I'm wrong about this we can see how to handle it for every browser independently.
- Usage of `navigator.platform.toLowerCase()` instead of reyling on the `OperatingSystem` type that is used everywhere else in the code.
- Unsigned commits => please read the contribution guideline.

Nevertheless, thanks for putting time and efforts into this. I will look into this feature request in the near future and will try to implement it with your approach as a reference.

PR #1472, opened 2025-10-29
feat(i18n): add zh-TW localization support

lynda0214:
Traditional Chinese wordings are quite different from Simplified Chinese. For the sake of user experience, I added the `zh-TW` translations into this project.


oliverschwendener (approved):
Lovely thanks

PR #1506, opened 2026-04-03
Added Quick Formatter

MarcoSennHaag:
Added new extension to quickly format XML, JSON and StackTraces. Also in a deep mode, which handles XML inside XML, JSON inside JSON and also JSON and XML inside StackTraces, etc.
Really helpful if debugging stuff as software developer.


src/main/Extensions/QuickFormatter/QuickFormatter.ts
```diff
@@ -0,0 +1,348 @@
+import { XMLBuilder, XMLParser } from "fast-xml-parser";
```
oliverschwendener: Could you use the `XmlParser` module from `src/main/Core/XmlParser` instead?

MarcoSennHaag: It's the same package, so I guess I could use it, but have to extend it. What's the benefit? External library just in one place?

oliverschwendener: If we ever want to replace `fast-xml-parser` with some other library or an own implementation we can do it in one place without touching any extensions.

oliverschwendener:
It would be nice to have the changes from `src/main/Core` in a separate PR.

scomans (dismissed):

MarcoSennHaag:
> It would be nice to have the changes from `src/main/Core` in a separate PR.

I can split the PR (once all issues are resolved)

MarcoSennHaag:
Please check #1508 first and I will update this one afterwards

src/main/Extensions/QuickFormatter/QuickFormatter.ts
```diff
@@ -0,0 +1,348 @@
...
+
+export class QuickFormatter {
+    public constructor(
+        private readonly xmlBuilder: XmlBuilder,
+        private readonly xmlBuilderOptions = {
+            format: true,
+            indentBy: "  ",
+            ignoreAttributes: false,
+            processEntities: true,
+        },
```
oliverschwendener: What's the reason for this to be configurable if we always pass the same options? https://github.com/MarcoSennHaag/ueli/blob/d1f462e6b8cf4323ff7ce29c3655a21c17be3198/src/main/Extensions/QuickFormatter/QuickFormatterExtension.ts#L151-L165

I suggest removing this argument from the constructor and making it a static variable inside this class.

src/main/Extensions/QuickFormatter/QuickFormatterExtension.ts
```diff
@@ -0,0 +1,177 @@
...
+                copyFormattedTextToClipboard: "Formatierten Text in die Zwischenablage kopieren",
+            },
+        };
+    }
+
+    private formatText(text: string, mode: "auto" | "stacktrace" | "json" | "xml"): string {
+        const enableDeepFormatting = this.getSettingValue("enableDeepFormatting");
+
+        const quickFormatter = new QuickFormatter(
+            new XmlBuilder(),
```
oliverschwendener: You're bypassing here the `XmlBuilder` module's contracts. You should get the xml builder from the module registry, pass it into this class as a dependency. Or even better pass in the `QuickFormatter` as a dependency into this class.

src/main/Extensions/QuickFormatter/QuickFormatter.ts
MarcoSennHaag: Enhanced

src/main/Extensions/QuickFormatter/QuickFormatterExtension.ts
MarcoSennHaag: You're right. Enhanced!

.github/CODEOWNERS
```diff
@@ -1,4 +1,5 @@
 # General stuff
+
```
oliverschwendener, line 2: These additional empty lines are not really necessary for your extension. Next time please avoid such changes.

oliverschwendener (approved):

.github/CODEOWNERS
MarcoSennHaag: I think this was auto-formatted by npm run check 🤔

oliverschwendener: `npm run check` does not perform any formatting, it only checks if the formatting is correct. And the code owners file is not included in the format check.

PR #1508, opened 2026-04-06
Added XmlBuilder and more parameters to XmlParser

MarcoSennHaag:
Added XmlBuilder and more parameters to XmlParser

### DIFF
--- a/src/main/Core/XmlBuilder/Contract/XmlBuilder.ts
+++ b/src/main/Core/XmlBuilder/Contract/XmlBuilder.ts
@@ -12,8 +12,8 @@ export interface XmlBuilder {
      * @param options.processEntities - Whether to process entities in the XML.
      * @returns The built XML string.
      */
-    build(
-        obj: unknown,
+    build<T extends object>(
+        obj: T,
         options?: { format: boolean; indentBy: string; ignoreAttributes: boolean; processEntities: boolean },
     ): string;
 }
--- a/src/main/Core/XmlBuilder/Contract/XmlBuilder.ts
+++ b/src/main/Core/XmlBuilder/Contract/XmlBuilder.ts
@@ -1,0 +1,19 @@
+/**
+ * Offers a method to build XML content.
+ */
+export interface XmlBuilder {
+    /**
+     * Builds XML from the given object.
+     * @param obj - The object to be converted to XML.
+     * @param [options] - Optional settings for building.
+     * @param options.format - Whether to format the output.
+     * @param options.indentBy - The string to use for indentation.
+     * @param options.ignoreAttributes - Whether to ignore attributes in the XML.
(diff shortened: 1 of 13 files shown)


scomans (dismissed):

src/main/Core/XmlBuilder/XmlBuilder.test.ts
```diff
@@ -0,0 +1,19 @@
+import { describe, expect, it } from "vitest";
+import { XmlBuilder } from "./XmlBuilder";
+
+describe(XmlBuilder, () => {
+    describe(XmlBuilder.prototype.build, () => {
+        it("should build and format XML file content", async () => {
```
oliverschwendener: `async` is not needed here as there is no `await` in the following code block.


```suggestion
        it("should build and format XML file content", () => {
```

src/main/Core/XmlBuilder/Contract/XmlBuilder.ts
```diff
@@ -0,0 +1,19 @@
...
+     * @param options.format - Whether to format the output.
+     * @param options.indentBy - The string to use for indentation.
+     * @param options.ignoreAttributes - Whether to ignore attributes in the XML.
+     * @param options.processEntities - Whether to process entities in the XML.
+     * @returns The built XML string.
+     */
+    build(
+        obj: unknown,
+        options?: { format: boolean; indentBy: string; ignoreAttributes: boolean; processEntities: boolean },
+    ): string;
```
oliverschwendener: Just wondering: whould it help this would be generic? Something like this?


```suggestion
    build<T>(
        obj: T,
        options?: { format: boolean; indentBy: string; ignoreAttributes: boolean; processEntities: boolean },
    ): string;
```

src/main/Core/XmlBuilder/XmlBuilder.test.ts
MarcoSennHaag: Done

src/main/Core/XmlBuilder/Contract/XmlBuilder.ts
MarcoSennHaag: Done

oliverschwendener: The change needs to be applied here in the contract as well.

By the way, what do you think about enforcing the type to be an object in order to prevent primitive types? Something like this?


```suggestion
    build<T extends object>(
        obj: T,
        options?: { format: boolean; indentBy: string; ignoreAttributes: boolean; processEntities: boolean },
    ): string;
```

Image

MarcoSennHaag: I changed it here, but this has some effects on my formatter extension in the other PR. But that's a second step.

oliverschwendener (approved):

PR #1526, opened 2026-06-18
VSCode: read recents from shared storage on 1.118+

IrishBruse:
### Summary

VS Code 1.118+ moved local recents into shared storage, so ueli was missing recently opened folders and files on newer versions. This PR merges recents from shared and user profile databases in order and deduplicates by URI.

- **Changed:** Recents database reads
- **Changed:** Command setting layout

## VS Code recents

Recents are loaded from shared storage, user profile keys, and legacy history entries, then combined in source order. Duplicate URIs are skipped, including `.code-workspace` config paths. Missing databases and read errors return empty results with logging, so older VS Code installs without shared storage still work.

## Settings

The command field tooltip is shown as a setting description instead of a separate info icon.

## Screenshots

Fixed on newer VS Code versions
image

Settings used to have a i circle
image

### DIFF
--- a/src/main/Extensions/VSCode/VSCodeExtension.ts
+++ b/src/main/Extensions/VSCode/VSCodeExtension.ts
@@ -95,13 +95,6 @@ export class VSCodeExtension implements Extension {
     private getRecentsRaw(): VSCodeRecentRaw[] {
         const userDatabasePath = this.stateDatabasePaths[this.operatingSystem];
 
-        if (this.useLegacyRecentsQuery()) {
-            return mergeRecents(
-                this.getRecentsFromDatabase(userDatabasePath, "recently.opened"),
-                this.getRecentsFromDatabase(userDatabasePath, "history.recentlyOpenedPathsList"),
-            );
-        }
-
         return mergeRecents(
             this.getRecentsFromDatabase(
                 this.sharedStateDatabasePaths[this.operatingSystem],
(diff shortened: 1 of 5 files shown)


src/main/Extensions/VSCode/VSCodeExtension.ts
```diff
@@ -86,18 +93,40 @@ export class VSCodeExtension implements Extension {
...
+    private getRecentsFromDatabase(databasePath: string, key: string): VSCodeRecentRaw[] {
+        if (!this.fileSystemUtility.existsSync(databasePath)) {
+            return [];
+        }
+
+        try {
+            const raw = new Database(databasePath, { readonly: true })
+                .prepare("SELECT json_extract(value, '$.entries') as entries FROM ItemTable WHERE key = ?")
+                .pluck()
+                .get(key) as string | null;
```
oliverschwendener: Just wondering what you think about this? I think if the function is offering generic type parameters, we should use them.

```suggestion
            const raw = new Database(databasePath, { readonly: true })
                .prepare<string, string | null>("SELECT json_extract(value, '$.entries') as entries FROM ItemTable WHERE key = ?")
                .pluck()
                .get(key);
```

src/main/Extensions/VSCode/VSCodeExtension.ts
```diff
@@ -57,6 +57,13 @@ export class VSCodeExtension implements Extension {
         Linux: process.env.HOME + "/.config/Code/User/globalStorage/state.vscdb",
     };
 
+    // VS Code 1.118+ stores local recents in application-shared storage.
+    private readonly sharedStateDatabasePaths = {
+        Windows: process.env.USERPROFILE + "/.vscode-shared/sharedStorage/state.vscdb",
```
oliverschwendener, line 62: Doesn't Windows use backslashes for file/folder paths? I'm not sure if this will work. Have you tested it?

IrishBruse: Node doesnt care you can mix them and generally / are prefered for cross platform as far as I know.

oliverschwendener (approved):

PR #1519, opened 2026-05-13
Add additional action to open folder

MarcoSennHaag:
First attempt for opening folders if file is found. Let me know if this suits your idea. Potentially handles #1512 
Expect some delays from my side the next days

PR #1527, opened 2026-06-19
Add additional actions to file search

MarcoSennHaag:
Signed commits from #1519

### DIFF
--- a/src/renderer/Core/Components/BasicSearch.tsx
+++ b/src/renderer/Core/Components/BasicSearch.tsx
@@ -93,9 +93,9 @@ export const BasicSearch = ({
             }
         };
 
-        window.addEventListener("keydown", windowKeyDownEventHandler as EventListener);
+        window.addEventListener("keydown", windowKeyDownEventHandler);
 
-        return () => window.removeEventListener("keydown", windowKeyDownEventHandler as EventListener);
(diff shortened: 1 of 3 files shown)


src/renderer/Core/Components/BasicSearch.tsx
```diff
@@ -46,21 +54,108 @@ export const BasicSearch = ({
...
+        const windowKeyDownEventHandler = (event: KeyboardEvent): void => {
+            if ((isMacOS ? event.metaKey : event.ctrlKey) && event.key.toLowerCase() === "k") {
+                event.preventDefault();
+                openAdditionalActionsMenu();
+            }
+        };
+
+        window.addEventListener("keydown", windowKeyDownEventHandler as EventListener);
+
+        return () => window.removeEventListener("keydown", windowKeyDownEventHandler as EventListener);
```
oliverschwendener: I just realized, these type casts are not necessary:


```suggestion
        window.addEventListener("keydown", windowKeyDownEventHandler);

        return () => window.removeEventListener("keydown", windowKeyDownEventHandler);
```

MarcoSennHaag: You are right. I removed them.

oliverschwendener (approved):

